PR #1, opened 2022-05-02
WIP Integrate current status of react-geo-client-template

hblitza:
Integrates latest react-geo-client-template ([status](https://github.com/terrestris/react-geo-client-template/commit/1c1e10ae0791eee2d85269d08003b99bab16d969)) to project.


dnlkoch:
IMHO the client should not contain the full history of the template client.

PR #6, opened 2022-05-06
Change eo layer

hblitza:
Change `demo eo layer`, since the old ones were very slow and sometimes unavailable.  

Might be enhanced later on.  

@terrestris/devs please check.


dnlkoch:
The e2e test needs to be adjusted accordingly.

hblitza:
Thanks

PR #24, opened 2022-05-16
Demo client login

FritzHoing:
A Header is created and styled using .less. 
Adds a Login-Button that directs the User to the login-Page.
PNG-Files can now be imported.


dnlkoch:
The `Header` and `Footer` elements were added in #41, but the `Login` button might be added in a separate MR.

dnlkoch:
Login is implemented in #126.

PR #41, opened 2022-05-19
Init Header and Footer

dnlkoch:
This introduces the `Header` and `Footer` components for the client:

Screenshot from 2022-05-19 18-13-34


src/variables.less
```diff
@@ -0,0 +1,4 @@
+@headerHeight: 50px;
```
KaiVolland, line 1: IMHO we should use css custom properties instead of less variables as they don't need import statements and can be modified at runtime.

KaiVolland (approved):

dnlkoch:
Thanks for the review!

PR #61, opened 2022-05-24
Introduce client configuration file

dnlkoch:
This introduces a configuration file for the client that can be used to set global options (currently the `appPrefix` is available only).

Please review @KaiVolland.


resources/public/fallbackConfig.js
```diff
@@ -0,0 +1,3 @@
+var clientConfig = {
```
KaiVolland, line 1: Wrong filename? comapre index.ejs

PR #66, opened 2022-05-25
Fix: broken assets path gh-pages

hblitza:
This adapts the base path in index.ejs template before building the gh-pages.  
Other Alternatives would be, e.g.
  - extra webpack-common for gh-pages

What do you think @terrestris/devs ?

### DIFF
--- a/.github/workflows/on-release.yml
+++ b/.github/workflows/on-release.yml
@@ -28,6 +28,9 @@ jobs:
     - name: Install dependencies ⏬
       run: npm ci
 
+    - name: Adapt base path for gh-pages :twisted_rightwards_arrows:
+      run: sed -i 's/\htmlWebpackPlugin.options.appPrefix/\/latest\//' resources/public/index.ejs
+
     - name: Build artifacts 🏗️
       run: npm run build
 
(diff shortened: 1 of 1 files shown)


.github/workflows/on-release.yml
```diff
@@ -28,6 +28,9 @@ jobs:
     - name: Install dependencies ⏬
       run: npm ci
 
+    - name: Adapt base path for gh-pages :twisted_rightwards_arrows:
```
dnlkoch, line 31:
```suggestion
    - name: Adapt base path for gh-pages 🔀
```

.github/workflows/on-release.yml
```diff
@@ -28,6 +28,9 @@ jobs:
...
+      run: sed -i 's/\htmlWebpackPlugin.options.appPrefix/\/latest\//' resources/public/index.ejs
```
dnlkoch, line 32: Basically +1, have you tested the updated path on the `gh-pages` branch beforehand?

dnlkoch, line 32: Basically +1, have you tested the updated path on the `gh-pages` branch beforehand?

dnlkoch:
@hblitza [Currently](https://terrestris.github.io/shogun-demo-client/latest/) it looks like the assets can be loaded without any issues. Can you confirm that? If so, is this fix still needed?

hblitza:
Yes, I can confirm this. Thanks, will close this.

PR #126, opened 2022-06-23
Init bearer token based authentication

dnlkoch:
This suggests to integrate the [keycloak-js](https://www.keycloak.org/docs/latest/securing_apps/index.html#_javascript_adapter) adapter for (bearer token based) authentication.

Due to this change the client is now capable to authenticate with a bearer token against the resource server (SHOGun, GeoServer and Print), this requires [PR 531](https://github.com/terrestris/shogun/pull/531) to be merged.

In addition the `UserMenu` is added containing the `ApplicationInfo`, logout and user account management entries.


LukasLohoff:
Very nice and the switch to tokens looks really useful! 

Did you add the Mapfish-Print serializers because direct authentication is now possible? At the first glance, these changes looked unrelated to authentication.

LukasLohoff (approved):

PR #172, opened 2022-07-11
feat: language selector

FritzHoing:
Implements a language selector into the item-container 'right items' of the header

ERM Ticket: https://redmine.intranet.terrestris.de/issues/14529

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -4,10 +4,9 @@ import {
   useAppSelector
 } from '../../hooks/useAppSelector';
 
-import i18n from '../../i18n';
 import BasicNominatimSearch from '../BasicNominatimSearch';
 
-import LanguageSelect from '../LanguageSelector/LanguageSelect';
+import LanguageSelect from '../LanguageSelector/Index';
 import UserMenu from '../UserMenu';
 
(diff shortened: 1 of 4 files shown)


src/components/LanguageSelector/LanguageSelect.tsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  return (
+    <div className="languageSelect">
+      <Select
+        defaultValue={i18n?.language || 'en'}
+        onChange={onLanguageChange}
+      >
+        <Option value="de">DE</Option>
+        <Option value="en">EN</Option>
```
dnlkoch, line 35: Not sure if possible, but I would prefer to fill the options dynamically with the provided locales from `i18n`.

src/components/Header/index.tsx
```diff
@@ -42,7 +45,8 @@ export const Header: React.FC<HeaderProps> = ({
       </div>
       <div
         className="item-container right-items"
-      >
+      > 
+        <LanguageSelect {...i18n} />
```
dnlkoch, line 49: `i18n` can be directly imported in the `LanguageSelect` instead of passing it as a prop.

PR #201, opened 2022-07-21
Harmonize button styles in measure and draw menu

dnlkoch:
This harmonizes the button styles in the measure and draw menu.

image

It also applies the style to the submenu that will be used for the collapsed tool menu.

image


Unraveler (approved):
Looks good to me

dnlkoch:
Thanks for the review!

PR #206, opened 2022-07-22
style: Toolmenu styling

FritzHoing:
Adds the library-implemented svg to point out that the layer hierarchy can be changed.

Auswahl_004


dnlkoch:
According to [the examples of antd](https://ant.design/components/tree/#components-tree-demo-draggable) this should already be available?

dnlkoch (approved):
Thanks @FritzHoing!

PR #208, opened 2022-07-25
Responsive layout

FritzHoing:
The language-selektor has been moved into the Toolbox. 

Layout for a Smartphone:
Auswahl_001
Layout for a TouchPad:
Auswahl_002

### DIFF
--- a/src/components/Header/index.less
+++ b/src/components/Header/index.less
@@ -60,11 +60,10 @@
     }
 
     .item-container {
-      display: flex;
 
       &.left-items {
         width: 60px;
-        max-width: 60px;
+        // max-width: 60px;
 
         .logo {
           width: 55px;
@@ -73,23 +72,23 @@
       }
 
       &.center-items {
-        align-items: center;
-        justify-content: center;
-        width: 100%;
+        // align-items: center;
+        // justify-content: center;
+        // width: 100%;
 
         .ant-select.ant-select-auto-complete.react-geo-nominatimsearch.ant-select-single.ant-select-allow-clear.ant-select-show-search {
           width: 250px;
 
           .ant-select-selector {
-            width: 100%;
+            // width: 100%;
           }
         }
       }
 
       &.right-items {
-        width: 60px;
+        // width: 60px;
         max-width: 60px;
-        align-items: center;
+        // align-items: center;
         justify-content: center;
 
         .username {
@@ -111,11 +110,11 @@
     }
 
     .item-container.center-items {
-      width: 100%;
+      // width: 100%;
       justify-content: center;
 
       .ant-select.ant-select-auto-complete.react-geo-nominatimsearch.ant-select-single.ant-select-allow-clear.ant-select-show-search {
-        width: fit-content;
+        // width: fit-content;
       }
     }
 
--- a/src/components/Header/index.less
+++ b/src/components/Header/index.less
@@ -52,6 +52,7 @@
     }
   }
 
+  // Adjustments for Smartphones
   @media (max-width:599px) {
 
     .title {
@@ -98,6 +99,7 @@
     }
   }
 
+  // Adjustments for Tablets
   @media (min-width:600px) {
     .title {
       display: none;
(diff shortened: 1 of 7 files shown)


src/components/Header/index.less
```diff
@@ -50,22 +44,82 @@
...
-      }
     }
+
     &.right-items {
       flex: 1;
       justify-content: flex-end;
     }
+  }
+
+  @media (max-width:599px) {
```
LukasLohoff, line 55: is this for smartphones or tablets or both? I think a short comment explaining that would be useful here

Unraveler (approved):
Looks good to me

src/components/Header/index.less
```diff
@@ -50,22 +44,82 @@
...
+        width: fit-content;
+      }
+    }
+
+    .item-container.right-items {
+      width: 100px;
+      max-width: 100px;
+    }
+  }
+}
```
LukasLohoff, line 125: missing newline at end of file. Do you have the `editorconfig` plugin installed?

LukasLohoff: that should add these automatically, and also configure other editor settings for you (whitespace indentation, etc.)

src/components/LanguageSelector/index.less
```diff
@@ -0,0 +1,4 @@
+.languageSelect {
```
LukasLohoff, line 1:
```suggestion
.language-select {
```

LukasLohoff: we usually format css classes like this. if you accept this, you'll also have to adjust the `LanguageSelect` component though

LukasLohoff (changes requested):
Nice! I suggested some small changes, maybe you can have a look at the comments below.

LukasLohoff (approved):

src/components/LanguageSelector/index.less
```diff
@@ -0,0 +1,9 @@
+.language-select {
+  display: flex;
+  justify-content: center;
+
+  .ant-select {
+    width: 100%;
+    margin: 5%;
```
dnlkoch, line 7: Is `5%` intended here?

src/components/Header/index.tsx
```diff
@@ -45,7 +45,6 @@ export const Header: React.FC<HeaderProps> = ({
       <div
         className="item-container right-items"
       >
-        <LanguageSelect />
```
dnlkoch, line 48: Can we remove the import as well?

src/components/LanguageSelector/index.less
```diff
@@ -0,0 +1,9 @@
+.language-select {
+  display: flex;
+  justify-content: center;
```
dnlkoch, line 3: Is this really needed?

src/components/ToolMenu/index.tsx
```diff
@@ -304,6 +306,22 @@ export const ToolMenu: React.FC<ToolMenuProps> = ({
...
+      label: <Menu.Divider />
+    },
+    {
+      key: 'LanguageSelect',
+      onTitleClick: onSubmenuTitleClick,
+      icon: <FontAwesomeIcon icon={faLanguage} />,
+      label: t('ToolMenu.languageSelect'),
+      children: [
+        {
+          key: 'LanguageSelect',
```
dnlkoch, line 320: The key is used in the surrounding block as well.

src/components/LanguageSelector/index.less
FritzHoing: Yes, this is the margin of the dropdown-menu:
Auswahl_003

dnlkoch: Yes, but why it's relative?

src/components/LanguageSelector/index.less
FritzHoing: Indeed, justify-content can be removed. But display: flex is needed. If it is missing, the drop down-menu is not completely within the toolbox.

src/components/LanguageSelector/index.less
FritzHoing: You are right! An absolute margin value makes more sense.

weskamm:
:heart:

dnlkoch (approved):

PR #249, opened 2022-08-10
fix: assigns the correct name to identify the layer

FritzHoing:
The WMS-layers could not be removed because the variable targetFolderName was assigned to a wrong name, so targetGroup was undefined.
Thus the unneeded key value pair externalWmsFolder: 'External services' was deleted.


LukasLohoff (approved):

Unraveler (approved):
Looks good to me

PR #250, opened 2022-08-10
Support plugins

dnlkoch:
This draft shows how we could add support for plugins to the shogun-client. Currently the following features are supported:

- Integration into the existing development setup (hot reloading, TypeScript support and others)
- Support for (nested) components
- Support for styling
- Support for i18n
- Support for own reducers and actions in the existing store

image

TODOs:

- Actually it's possible to integrate the plugins to the `ToolMenu` only, other integration points as a map control or a footer element should be implemented
- The production build is generally working, but not tested
- Documentation

Please note @terrestris/devs.


hwbllmnn (approved):
Very nice!

ahennr (approved):
:top: 👏🏼

dnlkoch:
Thanks for the reviews!

PR #286, opened 2022-08-24
Bugfix permalink

FritzHoing:
Permalink gets now also created out of ImageLayers.
The component refreshes automatically now since the state gets updated every time the zoomlevel, layer-composition or resolution changes.

### DIFF
--- a/src/components/Permalink/index.tsx
+++ b/src/components/Permalink/index.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import React, {
+  useEffect, useState
+} from 'react';
 
 import {
   CopyOutlined,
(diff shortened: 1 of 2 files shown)


src/components/Permalink/index.tsx
```diff
@@ -28,6 +40,8 @@ export interface PermalinkProps extends Partial<DefaultPermalinkProps> { }
 
 export const Permalink: React.FC<PermalinkProps> = () => {
 
+  const [permalink, setPermalink] = useState('');
```
dnlkoch, line 43: This will result in an empty link on initial render. We might want to call `PermalinkUtil.getLink` instead of providing an empty string.

dnlkoch (approved):
Very nice @FritzHoing!

PR #355, opened 2022-09-21
Permalink bugfixes

LukasLohoff:
- fixes listener that updates the permalink when layers are added / removed from layergroups
- refactors hook dependencies and introduces rules-for-hooks via `@terrestris/eslint-config-typescript-react`
- contains a bugfix which creates the processed layer group in bootstrap.tsx in the default application. thanks @FritzHoing 

### DIFF
--- a/src/components/BasicMapComponent/index.tsx
+++ b/src/components/BasicMapComponent/index.tsx
@@ -1,5 +1,6 @@
 import React, {
-  useEffect
+  useEffect,
+  useCallback
 } from 'react';
 
 import _isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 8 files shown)


src/components/Permalink/index.tsx
```diff
@@ -40,85 +41,95 @@ export interface PermalinkProps extends Partial<DefaultPermalinkProps> { }
...
+  const initialPermalink = map ? PermalinkUtil.getLink(
     map,
     ';',
     l => l.get('name'),
     l => (l instanceof TileLayer || l instanceof ImageLayer) && l.getVisible(),
     layerAttributes
-  ));
+  ) : '';
+
+  const [permalink, setPermalink] = useState(initialPermalink);
```
FritzHoing, line 57: That's a nice idea!

src/components/BasicMapComponent/index.tsx
```diff
@@ -79,7 +86,7 @@ export const BasicMapComponent: React.FC<Partial<MapComponentProps>> = ({
     } catch (error) {
       Logger.error(error);
     }
-  };
+  }, [client, map, t]);
```
FritzHoing, line 89: Is the translation function added here in case the translation - and thus the name of the layer group - changes?

FritzHoing (approved):
Very nice @LukasLohoff !

src/components/BasicMapComponent/index.tsx
LukasLohoff: yes, this is so the hook is called again if for example the translation function is not available in the first render

see https://reactjs.org/docs/hooks-rules.html

PR #356, opened 2022-09-21
feat: all imported layers can be deleted now

FritzHoing:
title says it all

### DIFF
--- a/src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
+++ b/src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
@@ -45,7 +45,6 @@ import {
 import Logger from '@terrestris/base-util/dist/Logger';
 
 import LayerUtil from '@terrestris/ol-util/dist/LayerUtil/LayerUtil';
-import MapUtil from '@terrestris/ol-util/dist/MapUtil/MapUtil';
 
 import {
   useMap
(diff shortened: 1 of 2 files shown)


src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -123,19 +123,21 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
...
     }
 
-    const targetFolderName = t('AddLayerModal.externalWmsFolder');
-    const targetGroup: LayerGroup = MapUtil.getLayerByName(map, targetFolderName) as LayerGroup;
-    if (targetGroup) {
-      const existingLayers = targetGroup.getLayers();
-      try {
-        if (existingLayers.getLength() === 1) {
-          targetGroup.set('hideInLayerTree', true);
+    const layerGroups = map.getLayers().getArray();
```
annarieger, line 126: This will match all map layers (not only imported and/or external). Is this intended?

FritzHoing: Yes, because we do not know yet if another layergroup will be added in the future. That's why there is the check in line 162, that looks for the attribute 'isImportedLayer'.

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -123,19 +123,21 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
...
+
+    layerGroups.forEach(layerGroup => {
+      if (layerGroup instanceof LayerGroup) {
```
LukasLohoff, line 129: is this `instanceof` check safe here?

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
LukasLohoff: agreed, I think there is no other way besides going through each layer in the map.

note that this will just give you layers on the first level, and we only support "custom" layerGroups on this level.

LukasLohoff (approved):

FritzHoing:
Thank you for the detailed review @annarieger and @LukasLohoff!

PR #370, opened 2022-09-28
feat: generates a new permalink whenever a layer is added

FritzHoing:
A new permalink is generated every time the zoom-level, the extent or the layer constellation changes.
This MR removes all project specific components out of the GIS-client and let's the GIS-client handle new imports in a generalized way.


LukasLohoff (approved):
Looks good! I think there is one empty return block (see comment). Otherwise just minor linting suggestions.

PR #376, opened 2022-09-29
Add print url to client config

Unraveler:
This makes the url to the print servlet configurable via the `gis-client-config.js`


LukasLohoff (approved):
:+1:

PR #400, opened 2022-10-10
fix: save opacity in permalink

LukasLohoff:
- saves opacity in permalink
- adds required listeners to keep peramlink updated
- increase clickable area of permalink icon

### DIFF
--- a/src/components/Permalink/index.tsx
+++ b/src/components/Permalink/index.tsx
@@ -40,7 +40,7 @@ export interface PermalinkProps extends Partial<DefaultPermalinkProps> { }
 
 export const Permalink: React.FC<PermalinkProps> = () => {
   const map = useMap();
-  const layerAttributes = useMemo(() => ['layerConfig', 'isExternalLayer', 'groupName', 'opacity'], []);
+  const layerAttributes = useMemo(() => ['layerConfig', 'isExternalLayer', 'groupName'], []);
   const {
     t
   } = useTranslation();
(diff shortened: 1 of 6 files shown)


Unraveler (approved):

src/components/Permalink/index.tsx
```diff
@@ -89,6 +89,18 @@ export const Permalink: React.FC<PermalinkProps> = () => {
       );
     };
 
+    const updateLayerConfig = () => {
+      const externalLayers = map.getAllLayers().filter(l => l.get('isExternalLayer'));
```
dnlkoch, line 93: Why do we restore the opacity for external layers only?

LukasLohoff: good point!

right now, because we only store the entire layer config for external layers. This requires a small refactoring though to harmonize `layers` and `customLayerParams`.

LukasLohoff: I'll do this in a separate PR, because I have more ideas for a refactoring. generally my idea would be:

- use only one URL parameter to store layer information for both external and persisted layers
   - this includes visibility, opacity, isExternalLayer, groupName etc. 
- move `restoreTransientLayers` from here to PermalinkUtil
- if possible, move PermalinkUtil to https://github.com/terrestris/shogun-util since the permalink util functions are more closely related to SHOGun application state than to openlayers (in my opinion)

PR #409, opened 2022-10-13
Draft: add download layer function

LukasLohoff:


src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -139,6 +159,28 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
...
   };
 
+  const downloadLayer = (url: string) => {
+    if (!layer) {
+      return;
+    }
+    const reqOpts = {
+      method: 'GET',
+      headers: {
+        ...getBearerTokenHeader(client?.getKeycloak())
```
dnlkoch, line 169: This should be applied to layers that are configured to do so only.

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -139,6 +159,28 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
...
+    fetch(url, reqOpts)
+      .then((res) => {
+        return res.blob();
+      })
+      .then((blob) => {
+        const a = document.createElement('a');
+        a.href = URL.createObjectURL(blob);
+        a.setAttribute('download', layer.get('name'));
+        a.click();
+      });
```
dnlkoch, line 181: Minor, but can we switch to `async/await` instead?

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -165,6 +207,18 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
     });
   }
 
+  if (downloadConfig) {
+    const downloadItems = downloadConfig.map((dlConfig: DownloadConfig) => {
+      return {
+        label: t('LayerTreeContextMenu.downloadLayer', {
+          formatName: dlConfig.formatName
```
dnlkoch, line 214: Is there a default for the `formatName` if not given?

dnlkoch (approved):
Very cool feature @LukasLohoff!

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
LukasLohoff: When it's a geoserver the default will be xml, so we could use that. If it is for some reason not a geoserver or another external link, the user can still overwrite it.

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
LukasLohoff: yes, it is fixed in 7e07cc76c5179b2f01be452d57a32fe7458148d9

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
LukasLohoff: done, see 7e07cc76c5179b2f01be452d57a32fe7458148d9

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
LukasLohoff: see c3b9d8e8e1b217d50fa20dfe1c8fed84f459db36

LukasLohoff:
Thanks for the detailed review! For the XML default I had change how the download URL is determined in c3b9d8e8e1b217d50fa20dfe1c8fed84f459db36

PR #425, opened 2022-10-20
feat: add background layer switcher

hwbllmnn:
See title.  @terrestris/devs Please review.


src/App.less
```diff
@@ -24,3 +24,14 @@
...
+.bg-layer-chooser {
+  position: fixed;
+  bottom: 70px;
+  right: 30px;
+  z-index: 10;
+}
+
+.bg-preview {
+  z-index: 10;
+}
```
dnlkoch, line 37: Missing newline.

src/App.tsx
```diff
@@ -24,12 +29,19 @@ export const App: React.FC<AppProps> = ({
...
+    layers = map.getLayers().getArray();
+  }
+
   return (
     <div
       className="App"
       {...restProps}
     >
       <Header />
+      <BackgroundLayerChooser layers={layers as any} />
```
dnlkoch, line 44: This would add _all_ map layers to the chooser. I would favour to add a selection of intended (background) layers only.

dnlkoch (changes requested):

src/App.tsx
hwbllmnn: Not really, the [default layer filter](https://github.com/terrestris/react-geo/blob/main/src/BackgroundLayerChooser/BackgroundLayerChooser.tsx#L26) makes sure only layers with the `isBackgroundLayer` property are added.

hwbllmnn: BTW: is there a reason for the grouping in the `bootstrap.tsx`? I've made some changes there in order to be able to test this properly, which I didn't commit because I wasn't sure why that is.

dnlkoch: The `bootstrap.tsx` is the async boundary for the module federation plugin which "split[s] out the initialization code of a larger chunk to avoid any additional round trips and improve performance in general" (see [here](https://webpack.js.org/concepts/module-federation/) for more details).

dnlkoch (approved):

Issue #485, opened 2022-11-14
Background Layer Switcher is broken


dnlkoch:
The background layer switcher seems to be broken (at least in the hosted example app):

![image](https://user-images.githubusercontent.com/1137620/201632851-7de12485-5d51-436e-bc58-3ee3cf83df81.png)

PR #510, opened 2022-11-28
Disable the `BackgroundLayerChooser`

dnlkoch:
This suggest to disable the `BackgroundLayerChooser` as it seems to be broken right now (see #485) and:

- There is no way to define the required property `isBackgroundLayer` in shogun nor the shogun-admin.
- There seems to be a conflict by setting it to a layer that is also visible in the tree (the layer _must_ be visible in the tree to be catched up by the chooser).
- There seems to be a request loop once a layer is activated in it.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,11 +2,6 @@ import React, {
   useEffect
 } from 'react';
 
-import {
-  BackgroundLayerChooser,
-  useMap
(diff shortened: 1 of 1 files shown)


FritzHoing (approved):

dnlkoch:
Thanks for the review!

PR #522, opened 2022-12-01
Bugfixing

FritzHoing:
- Removes Line because of inconsistent behavior

### DIFF
--- a/src/components/PrintForm/Serializer/SHOGunMapFishPrintV3WMSSerializer.ts
+++ b/src/components/PrintForm/Serializer/SHOGunMapFishPrintV3WMSSerializer.ts
@@ -23,6 +23,10 @@ export class SHOGunMapFishPrintV3WMSSerializer extends MapFishPrintV3WMSSerializ
       serialized.baseURL = `http://shogun-geoserver:8080${serialized.baseURL}`;
     }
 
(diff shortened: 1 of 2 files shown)


src/components/PrintForm/Serializer/SHOGunMapFishPrintV3WMSSerializer.ts
```diff
@@ -23,6 +23,10 @@ export class SHOGunMapFishPrintV3WMSSerializer extends MapFishPrintV3WMSSerializ
       serialized.baseURL = `http://shogun-geoserver:8080${serialized.baseURL}`;
     }
 
+    if (serialized.baseURL.startsWith('/geoserver/SHOGUN')) {
+      serialized.baseURL = `http://shogun-geoserver:8080${serialized.baseURL}`;
```
dnlkoch, line 27: Even if the value is not intended to be changed, we might want to move the base url to the external config file.

dnlkoch (approved):

src/components/PrintForm/Serializer/SHOGunMapFishPrintV3WMSSerializer.ts
FritzHoing: Yes thank you for the idea @dnlkoch i will fix this in a follow up

PR #528, opened 2022-12-05
Tool menu refactoring

annarieger:
* ⚠️ BREAKING CHANGE ⚠️ Replace `Menu` by `Collapse`
*  Get rid of dirty hack while mapfish print initialization within a menu item, move logic to the `PrintForm`
*  Get rid of some antd deprecations
   *  See also https://github.com/terrestris/react-geo/pull/3006
* Replace less variables (`@`) by CSS variables

image

### DIFF
--- a/src/components/PrintForm/index.tsx
+++ b/src/components/PrintForm/index.tsx
@@ -1,4 +1,6 @@
 import React, {
+  useCallback,
+  useEffect,
   useState
 } from 'react';
 
@@ -19,6 +21,7 @@ import {
   FormProps
 } from 'antd/lib/form/Form';
 
+import ClientConfiguration from 'clientConfig';
 import OlLayerGroup from 'ol/layer/Group';
 import OlLayer from 'ol/layer/Layer';
 import OlMap from 'ol/Map';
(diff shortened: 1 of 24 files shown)


src/components/PrintForm/index.tsx
```diff
@@ -53,23 +69,84 @@ export interface PrintFormProps extends Omit<FormProps, 'form'> {
...
+    pManager.serializers = [
+      new MapFishPrintV3GeoJsonSerializer(),
+      new MapFishPrintV3OSMSerializer(),
+      new MapFishPrintV3WMTSSerializer(),
+      new SHOGunMapFishPrintV3WMSSerializer(),
+      new SHOGunMapFishPrintV3TiledWMSSerializer()
+    ];
+
+    await pManager.init()
+      .then(() => {
```
dnlkoch, line 127: Since we make use of `await` `then()` is not needed.

dnlkoch (approved):

src/components/PrintForm/index.tsx
annarieger: Nice (copypasted) catch! Fixed with [e20f4b5](https://github.com/terrestris/shogun-gis-client/pull/528/commits/e20f4b5c1e0fa094c9df89a14741fd9a97fb63cc)

PR #540, opened 2022-12-07
style: adjustments of the toolbar

FritzHoing:
- removes max hight of content box
- removes custom icon transform due to incompatible icons

### DIFF
--- a/src/components/ToolMenu/index.less
+++ b/src/components/ToolMenu/index.less
@@ -34,7 +34,7 @@
     }
   }
 
-  .ant-collapse {
+  >.ant-collapse {
     border: none;
     max-height: calc(100vh - var(--headerHeight) - var(--footerHeight) - 32px);
     overflow-y: auto;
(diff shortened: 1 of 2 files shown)


src/components/ToolMenu/index.less
```diff
@@ -52,7 +52,6 @@
       .ant-collapse-content-box {
         padding: 10px;
         overflow-y: auto;
-        max-height: 25vh;
```
annarieger, line 55: That makes the whole collapse scrollbar. The intention was to make the certain panels scrollable instead.

src/components/ToolMenu/index.tsx
```diff
@@ -292,7 +292,6 @@ export const ToolMenu: React.FC<ToolMenuProps> = ({
   return (
     <div className={`tool-menu ${collapsed ? 'collapsed' : ''}`}>
       <Collapse
-        expandIcon={() => <FontAwesomeIcon icon={faChevronDown} />}
```
annarieger, line 295: just wondering, why are the icons incompatible?

PR #596, opened 2023-01-03
fix(issue): disable measure tools when panel collapses, highlight active tool

weskamm:
If one closes the measure tools panel, its interactions will be disabled.

Besides, active measure and draw tools are highlighted a bit with a bigger font size.

image


hblitza (approved):

weskamm:
Thank you!

PR #600, opened 2023-01-04
Fix a conflict between measurement tools and the export (print) panel

mholthausen:
This PR fixes a conflict between the measurement tools and the export (print) panel where the measuring tools remain active when opening the print tool and thus led to a conflict when selecting the print frame and vice versa


hwbllmnn (approved):

mholthausen:
Subjected to extensive refactoring after review took place. Requesting for a new review of these changes @terrestris/devs

dnlkoch (approved):

PR #601, opened 2023-01-04
feat: add layer loading indicator to tree

weskamm:
This adds a loading indicator to layers to give a better user experience.

Peek 2023-01-04 15-52

Some layer sources, like OSM, may have a flickering effect as their loading state cannot always be determined completely.

### DIFF
--- a/src/components/ToolMenu/LayerTree/index.tsx
+++ b/src/components/ToolMenu/LayerTree/index.tsx
@@ -1,6 +1,9 @@
 import React, {
-  useState
+  useState,
+  useEffect
 } from 'react';
+
+import { Progress } from 'antd';
 
 import {
   getUid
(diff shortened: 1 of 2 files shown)


src/components/ToolMenu/LayerTree/index.tsx
```diff
@@ -55,6 +58,59 @@ export const LayerTree: React.FC<LayerTreeProps> = ({
   } = useTranslation();
 
   const [visibleLegendsIds, setVisibleLegendsIds] = useState<string[]>([]);
+  const [layerTileLoadCounter, setLayerTileLoadCounter] = useState<any>({});
```
dnlkoch, line 61: Is there a type for the counter available?

src/components/ToolMenu/LayerTree/index.tsx
```diff
@@ -55,6 +58,59 @@ export const LayerTree: React.FC<LayerTreeProps> = ({
...
+  useEffect(() => {
+    if (!map) {
+      return;
+    }
+    const allLayers = MapUtil.getAllLayers(map);
+    allLayers.forEach(layer => {
+      if (layer instanceof OlLayer) {
+        const source = layer.getSource();
+        if (!source.hasListener('tileloadstart')) {
+          source.on('tileloadstart', () => {
```
dnlkoch, line 72: Do we need to unregister the listeners on unmount?

dnlkoch (approved):
Very nice @weskamm!

src/components/ToolMenu/LayerTree/index.tsx
weskamm: good catch, added

PR #619, opened 2023-01-10
style: add hover style on some elements

annarieger:
Use `secondaryColor` as hover color on some elements like links in footer, menu item titles in main menu etc.

Non-functional change.


weskamm (approved):
A new variable `hoverColor` could have been used also, but i think its ok like this

PR #630, opened 2023-01-13
fix: adds a check for array length

FritzHoing:
Empty tool configurations will now be replaced with the full list of available tools.


weskamm (approved):
:+1:

PR #486, opened 2022-11-14
Adds a cookie-consent

FritzHoing:
A new cookie-consent is added.
If cookie-plugins are available, the written content can be received from the plugins.

Auswahl_046

PR #642, opened 2023-01-18
Introduce cookie consent

mholthausen:
This PR introduces a cookie banner consent and fixes some eslint code warnings.

https://github.com/terrestris/shogun-gis-client/pull/486 is closed in favor of this PR.

### DIFF
--- a/src/components/CookieBanner/index.less
+++ b/src/components/CookieBanner/index.less
@@ -1,9 +1,31 @@
 .cookie-container {
-  background: var(--primaryColor) !important;
-}
+  background: var(--primaryColor);
+  bottom: 0px;
+  align-items: baseline;
+  color: var(--complementaryColor);
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  left: 0px;
+  position: fixed;
+  width: 100%;
+  z-index: 999;
 
-.cookie-button {
-  position: relative;
-  color: var(--complementaryColor) !important;
-  background: var(--secondaryColor) !important;
+  .cookie-content {
+    flex: 1 0 300px;
+    margin: 15px;
+  }
+
+  .cookie-button {
+    position: relative;
+    color: var(--complementaryColor);
+    background: var(--secondaryColor);
+    border: 0px;
+    border-radius: 0px;
+    box-shadow: none;
+    cursor: pointer;
+    flex: 0 0 auto;
+    padding: 5px 10px;
+    margin: 15px;
+  }
 }
--- a/src/components/CookieBanner/index.less
+++ b/src/components/CookieBanner/index.less
@@ -1,0 +1,9 @@
+.cookie-container {
+  background: var(--primaryColor) !important;
+}
+
+.cookie-button {
+  position: relative;
+  color: var(--complementaryColor) !important;
+  background: var(--secondaryColor) !important;
+}
--- a/src/components/CookieBanner/index.tsx
+++ b/src/components/CookieBanner/index.tsx
@@ -30,21 +30,19 @@ export const CookieBanner = () => {
   };
 
   return (
-    <div className="CookieBanner">
-      <CookieConsent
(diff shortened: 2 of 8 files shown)


src/components/CookieBanner/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const CookieBanner = () => {
+  const plugins = usePlugins();
+
+  const { t } = useTranslation();
+
+  const getPlugin = () => {
+    let cookiePlugin: React.ReactElement | null = null;
+
+    // TODO: warn if more than one cookie-plugin has been detected
```
dnlkoch, line 19: Is this still relevant?

src/i18n/translations.ts
```diff
@@ -14,12 +14,19 @@ export default {
...
+          'Das Capabilities Dokument konnte nicht geladen werden, bitte prüfen Sie die URL',
         version: 'Version'
       },
       BasicMapComponent: {
         processedLayersFolder: 'Prozessierte Layer'
       },
+      CookieBanner: {
+        info:
+          'Wir verwenden Cookies, um Ihnen die bestmögliche Nutzung unserer Website zu ermöglichen. ' +
+          'Wenn Sie diese Website weiterhin nutzen, gehen wir davon aus, dass Sie damit zufrieden sind.',
```
dnlkoch, line 27: I assume we need to provide proper information about the used cookies.

src/i18n/translations.ts
```diff
@@ -230,10 +249,12 @@ export default {
       },
       Index: {
         applicationLoadErrorMessage: 'Error while loading the application',
-        applicationLoadErrorDescription: 'The application with ID {{applicationId}} could not be loaded correctly. ' +
+        applicationLoadErrorDescription:
+          'The application with ID {{applicationId}} could not be loaded correctly. ' +
```
FritzHoing, line 253:
```suggestion
          'The application with ID {{applicationId}} could not be loaded correctly. '
```

src/components/CookieBanner/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return cookiePlugin;
+  };
+
+  return (
+    <div className="CookieBanner">
+      <CookieConsent
+        key="default-cookie-consent"
+        location="bottom"
+        buttonText={t('CookieBanner.button')}
+        cookieName="SHOGun-Session-Cookie"
```
dnlkoch, line 40: Is this name used somewhere? If so, the backend doesn't make use of any session cookies and it should be adapted.

src/components/CookieBanner/index.less
```diff
@@ -0,0 +1,9 @@
+.cookie-container {
+  background: var(--primaryColor) !important;
+}
+
+.cookie-button {
+  position: relative;
+  color: var(--complementaryColor) !important;
+  background: var(--secondaryColor) !important;
+}
```
dnlkoch, line 9: Can we get rid of the `!important`s?

src/components/CookieBanner/index.tsx
mholthausen: As far as I see it, it's not used anywhere. It is intended that the user agrees to this if they continue to use the application. I don't think the backend needs to be modified to do this, otherwise the user will have no choice but to leave the page.  Otherwise we should discuss it further.

mholthausen: The cookie mentioned here is used exclusively in the client and is saved in the browser storage within the cookies area when accepted by the user so that this message is no longer displayed when the client is called up again (until the cookie has been removed or has expired). If the user does not agree, this message will prompt again each time the client is opened.

dnlkoch: Thanks for the clarification!

dnlkoch (approved):
Very nice, thanks @mholthausen !

PR #646, opened 2023-01-20
Use legal links config

mholthausen:
This PR lets the map client use the configuration of the legal links for the footer, when available. If this setting is not specified for a client, the already existing defined links are used as default.  
Needs https://github.com/terrestris/shogun/pull/642 and https://github.com/terrestris/shogun-util/pull/254 first be merged.

Also included is some code styling by eslint.

### DIFF
--- a/src/components/Footer/index.tsx
+++ b/src/components/Footer/index.tsx
@@ -1,21 +1,14 @@
-import React, {
-  useEffect
-} from 'react';
+import React, { useEffect } from 'react';
 
 import {
-  Button,
-  Divider
+  Button, Divider
 } from 'antd';
 
 import OlControlMousePosition from 'ol/control/MousePosition';
 import OlControlScaleLine from 'ol/control/ScaleLine';
-import {
-  createStringXY
-} from 'ol/coordinate';
+import { createStringXY } from 'ol/coordinate';
 
-import {
-  useTranslation
-} from 'react-i18next';
+import { useTranslation } from 'react-i18next';
 
 import ScaleCombo from '@terrestris/react-geo/dist/Field/ScaleCombo/ScaleCombo';
 import useMap from '@terrestris/react-geo/dist/Hook/useMap';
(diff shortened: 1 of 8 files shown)


src/store/legal/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  createSlice, PayloadAction
+} from '@reduxjs/toolkit';
+
+export type Legal = {
+  contact?: string;
+  imprint?: string;
+  privacy?: string;
+};
+
+const initialState: Legal = {};
```
dnlkoch, line 11: We might want to set the defaults here.

src/components/Footer/index.tsx
```diff
@@ -94,37 +102,19 @@ export const Footer: React.FC<FooterProps> = ({
...
-          />
+          <Divider type="vertical" />
         </div>
-        <div
-          className="mouse-position-wrapper"
-        >
+        <div className="mouse-position-wrapper">
           {t('Footer.mousePosition')}:&nbsp;
           <div
             id="mouse-position"
```
dnlkoch, line 120: I would favour the previously adapted code style, but this is non-blocking.

src/store/legal/index.ts
mholthausen: Defaults got moved from the component to the store.

src/components/Footer/index.tsx
```diff
@@ -1,44 +1,40 @@
...
-  useTranslation
-} from 'react-i18next';
+import { useTranslation } from 'react-i18next';
 
 import ScaleCombo from '@terrestris/react-geo/dist/Field/ScaleCombo/ScaleCombo';
 import useMap from '@terrestris/react-geo/dist/Hook/useMap';
 
 import './index.less';
+import useAppSelector from '../../hooks/useAppSelector';
+import { Legal } from '../../store/legal';
```
simonseyock, line 18: This looks like it could have linting issues, please check,

mholthausen: Checked it locally and all checks here run successfully. `.eslintrc.js` says curly new line warning starts at two properties. VS Code edited the entries with just one property automatically on save

PR #702, opened 2023-02-21
Add ScaleSelect to print form

annarieger:
* Introduce `ScaleSelect` component to print form which allows to choose print scale via dropdown
* Allow to pass an optional parameter `customPrintScales` to the `PrintForm` 
* Remove hardcoded values for initial DPI and output formats, use dynamic values set on mapfish print provider instead
* Get rid of some `@ts-ignore` annotations


src/components/PrintForm/ScaleSelect/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+export const ScaleSelect: React.FC<ScaleSelectProps> = ({
+  printManager,
+  placeholder,
+  value,
+  ...restProps
+}): JSX.Element => {
+
+  const map = useMap();
+
+  const [scale, setScale] = useState<any>(value);
```
dnlkoch, line 32: Any chance to get rid of the `any` type?

src/components/PrintForm/ScaleSelect/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  useEffect(() => {
+    if (printManager) {
+      printManager.setScale(scale);
+    }
+  }, [printManager, scale]);
+
+  useEffect(() => {
+
```
dnlkoch, line 41:
```suggestion
```

dnlkoch (approved):

src/components/PrintForm/ScaleSelect/index.tsx
annarieger: done with 88360d8

PR #706, opened 2023-02-23
Fix FeatureInfo for plugins

annarieger:
Pass `client` as property instead of getting it from context via hook.

This is needed for the case if the `FeatureInfo` component is used with some plugin, where the client context is not  directly accessible.

Also enhance logic to estimate query layers:
- exclude invisible layer candidates
- exclude WMTS layers

### DIFF
--- a/src/components/ToolMenu/FeatureInfo/index.tsx
+++ b/src/components/ToolMenu/FeatureInfo/index.tsx
@@ -12,6 +12,7 @@ import OlLayerImage from 'ol/layer/Image';
 import OlLayerTile from 'ol/layer/Tile';
 import OlSourceImageWMS from 'ol/source/ImageWMS';
 import OlSourceTileWMS from 'ol/source/TileWMS';
+import OlSourceWMTS from 'ol/source/WMTS';
 
 import {
   Tab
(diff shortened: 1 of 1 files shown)


src/components/ToolMenu/FeatureInfo/index.tsx
```diff
@@ -57,18 +59,19 @@ import './index.less';
...
   enabled,
+  client,
   ...restProps
 }): JSX.Element => {
   const {
     t
   } = useTranslation();
 
   const map = useMap();
-  const client = useSHOGunAPIClient();
```
dnlkoch, line 71: It also won't work if the plugin is wrapped with the appropriate context?

src/components/ToolMenu/FeatureInfo/index.tsx
```diff
@@ -112,10 +115,14 @@ export const FeatureInfo: React.FC<FeatureInfoProps> = ({
...
-
-          return unqualifiedLayerName === unqualifiedMapLayerName;
+          const source = l.getSource();
+          if (source instanceof OlSourceWMTS) {
+            return false;
+          } else {
+            const unqualifiedMapLayerName = getUnqualifiedLayerName(l.getSource().getParams().LAYERS);
+            const unqualifiedLayerName = getUnqualifiedLayerName(layerName);
+            return unqualifiedLayerName === unqualifiedMapLayerName;
+          }
```
dnlkoch, line 125: We might want to reuse the [isWmsLayer](https://github.com/terrestris/react-geo/blob/main/src/Util/typeUtils.ts#L10) guard from react-geo instead.

annarieger:
Replaced by #715 and #716

PR #789, opened 2023-04-13
Menu layout fixes

dnlkoch:
This fixes some minors in the menu layout if the overall width is increased by a plugin:

- always keep the menu title icon width the same
- Make use of the full width within the print form

Example:

image


hwbllmnn (approved):

dnlkoch:
Thanks for the review!

PR #885, opened 2023-05-31
Edit features via WFS-T

LukasLohoff:
This PR allows to create, edit and delete features via WFS-T.

image

image

- [x] Release & update shogun
- [x]  Release & update shogun-util
- [x] Ready for review

…


src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const onSaveSuccess = (responseText?: string) => {
+    if (!responseText) {
+      return;
+    }
+    setErrorMsg(undefined);
+    const parser = new DOMParser();
+    const xmlDoc = parser.parseFromString(responseText, 'text/xml');
+
+    // get feature id from response
+    const featureId = xmlDoc.getElementsByTagName('ogc:FeatureId');
```
ahennr, line 169: Is currently only WFS-T 2.0.0 supported?

src/hooks/useExecuteGetFeature.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (url.endsWith('?')) {
+      url = url.slice(0, -1);
+    }
+
+    const featureRequest = new OlFormatWFS().writeGetFeature({
+      srsName: map.getView().getProjection().getCode(),
+      featureNS: describeFeatureType.targetNamespace,
+      featurePrefix: describeFeatureType.targetPrefix,
+      featureTypes: [source?.getParams().LAYERS],
+      outputFormat: 'application/json',
```
ahennr, line 79: Will always GeoServer be used? Not all WFS servers support `application/json` as output…

ahennr (approved):
Awesone 👏🏼

src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
dnlkoch: WFS Transaction is currently implemented in [1.1.0](https://github.com/terrestris/shogun-gis-client/pull/885/files/0a016436a3a23be7586a12e5589fea6f61d2dd73#diff-c1019a11bf7b99a4475f13c98e1427accd8d2f03976f4071ca5553797c7eccf5R120).

src/hooks/useExecuteGetFeature.ts
dnlkoch: Internally GeoServer will always be used, but actually it's possible to include layers from external servers as well which might be used by the given method and might fail. Good point.

@LukasLohoff What do you think, should we switch to GML here instead?

src/i18n/translations.ts
```diff
@@ -167,6 +168,42 @@ export default {
         searchNominatim: 'Ortssuche (Nominatim)',
         nominatimTitle: 'Ortssuche',
         searchPlaceholder: 'Orts- und Datensuche…'
+      },
+      EditFeatureDrawer: {
+        featureEditor: 'Objekteditor',
+        noLayerFoundError: 'Kein passender Layer gefunden - das Editieren ist nicht möglich.',
+        isFeatureLockedErrorMsg: 'Das ausgewählte Objekt wird bereits in einer anderen Sitzung bearbeitet.',
+        closeDrawerWarnTitle: 'Wollen Sie wirklich fortfahren?',
+        closeDrawerWarnContent: 'Dadurch werden alle nicht gespeicherten Änderungen an dem Object verworfen.'
```
dnlkoch, line 177:
```suggestion
        closeDrawerWarnContent: 'Dadurch werden alle nicht gespeicherten Änderungen an dem Objekt verworfen.'
```

src/i18n/translations.ts
```diff
@@ -337,6 +375,42 @@ export default {
...
+      },
+      ResetButton: {
+        title: 'Reset'
+      },
+      SaveButton: {
+        title: 'Save'
+      },
+      DeleteButton: {
+        title: 'Delete object',
+        confirm: 'The object will be completely deleted. Proceed?'
```
dnlkoch, line 410:
```suggestion
        title: 'Delete feature',
        confirm: 'The feature will be completely deleted. Proceed?'
```

src/hooks/useExecuteGetFeature.ts
```diff
@@ -0,0 +1,106 @@
...
+import {
+  getBearerTokenHeader
+} from '@terrestris/shogun-util/dist/security/getBearerTokenHeader';
+
+import useExecuteWfsDescribeFeatureType from './useExecuteWfsDescribeFeatureType';
+import useSHOGunAPIClient from './useSHOGunAPIClient';
+
+export type GetFeatureOpts = {
+  layer: WmsLayer;
+  // TODO Pass filter instead of featureId
```
dnlkoch, line 31: This TODO should be fixed before merging.

src/components/MultiSearch/index.tsx
```diff
@@ -350,6 +383,43 @@ export const MultiSearch: React.FC<MultiSearchProps> = ({
...
+      dispatch(setLayerId(getUid(layer)));
+      dispatch(showEditFeatureDrawer());
+      setResultsVisible(false);
+    };
+
+    if (
+      allowedEditMode.includes('CREATE') ||
+      allowedEditMode.includes('DELETE') ||
+      allowedEditMode.includes('UPDATE')
+    ) {
```
dnlkoch, line 404: We should also check if the layer is marked as editable.

src/components/EditFeatureDrawer/EditFeatureForm/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+            {...fieldCfg?.fieldProps}
+          />
+        );
+      case 'TEXTAREA':
+        return (
+          <Input.TextArea
+            {...fieldCfg?.fieldProps}
+          />
+        );
+      // TODO Still under development, does it make sense anyway?
```
dnlkoch, line 151: Currently this field is not usable at all, I suggest to remove it.

src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+    const updatedFeatures = await executeGetFeature({
+      layer: layer,
+      featureId: id
+    });
+
+    if (
+      updatedFeatures?.features[0] &&
+      (allowedEditMode.includes('UPDATE') ||
+      allowedEditMode.includes('DELETE'))
```
dnlkoch, line 97: Why do we check for the roles here? I would assume it needs to be done _before_ the user actually can trigger updating/deleting a feature?

src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+              tooltip={t('EditFeatureGeometryToolbar.edit')}
+              onModifyStart={updateRevision}
+              onModifyEnd={updateRevision}
+              onTranslateEnd={updateRevision}
+              {...btnTooltipProps}
+            />
+            : <></>
+        }
+        {
+          allowedEditMode.includes('DELETE') ?
```
dnlkoch, line 241: This is actually refers to the deletion of the _geometry_ not the _feature_. Is this intended?

src/hooks/useExecuteGetFeature.ts
dnlkoch: Fixed in d2ea593.

src/hooks/useExecuteGetFeature.ts
LukasLohoff: good point @ahennr 
we'll test if we can switch to GML here

src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
mholthausen: Yes, the requirement was that in this case both form editing and geometry editing tools should be restricted according to the configuration.

src/components/MultiSearch/index.tsx
mholthausen: It's checked already [here](https://github.com/terrestris/shogun-gis-client/pull/885/files/12858ce4a26c9006c0f9ec6d0860af61981a29dc#diff-2d094e5695965d7155bb73a7f9398c30472ceb99f1de5478fc74a29fa1fed5e0R390) in  `actionsCreator` and will return if its not editable.

src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
dnlkoch: Alright, thanks!

src/hooks/useExecuteGetFeature.ts
LukasLohoff: we had some issues parsing the GML with openlayers. For now, we'll use JSON and may implement this in the future

src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
mholthausen: The check is obsolete at this point and has been removed.

PR #907, opened 2023-06-07
Configurable WFS LockFeature

dnlkoch:
This makes the usage of WFS LockFeature during feature editing sessions configurable (disabled by default).


LukasLohoff (approved):

dnlkoch:
Thanks for the review!

PR #927, opened 2023-06-13
Use custom print scale from map scales

mholthausen:
This PR uses `customPrintScales` from the defined map scales to keep the print scales and map scales in sync.

### DIFF
--- a/src/components/PrintForm/index.tsx
+++ b/src/components/PrintForm/index.tsx
@@ -130,8 +130,9 @@ export const PrintForm: React.FC<PrintFormProps> = ({
         ?.getView()
         ?.getResolutions()
         ?.map((d: number | undefined) => {
+          const units = map?.getView()?.getProjection()?.getUnits();
           if (typeof d !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/components/PrintForm/index.tsx
```diff
@@ -126,6 +126,20 @@ export const PrintForm: React.FC<PrintFormProps> = ({
     let pManagerOpts: MapFishPrintV3ManagerOpts = {
       url: ClientConfiguration.print?.url || '/print',
       map,
+      customPrintScales: map
+        ?.getView()
+        ?.getResolutions()
+        ?.map((d: number | undefined) => {
+          if (typeof d !== 'undefined') {
+            const scale = MapUtil.getScaleForResolution(d, 'm');
```
dnlkoch, line 134: What happens if the map units are not in meters?

dnlkoch (approved):

src/components/PrintForm/index.tsx
mholthausen: An automatic detection of map units has been added with 64fb597e54a7b786a9276f9208a7f80339fd1fd2.

PR #967, opened 2023-06-27
feat: configurable search result display

LukasLohoff:
- consider `searchConfig.displayTemplate` when displaying search results
- when displayTemplate is not set: enables solr highlighting, which allows us to highlight the search result even if there are no exact matches 
- removes the `listPrefixRenderer` for now, because the same info is already displayed in the collapse title
- changes the solr parameters to the keys used by solr to make the search endpoint interchangeable


dnlkoch (approved):
Very nice @LukasLohoff!

LukasLohoff:
Thanks for the review!

PR #976, opened 2023-06-29
Edit feature improvements

LukasLohoff:
- prevents overlapping property labels 
   - uses a default of 450px, configured via `--editDrawerWidth`
   - uses `word-break: break-word;` as a fallback, in case the label is too long and does not contain whitespaces
- determines whether form is dirty and save it to the redux store (`state.editFeature.formDirty`)
   - `SaveButton` is disabled if form is not diry 
- fixes `SaveButton` loading animation
- introduces `FeedbackIcon` which allows to show positive feedback after saving features:

save-success


src/components/EditFeatureDrawer/SaveButton/index.tsx
```diff
@@ -115,21 +130,29 @@ export const SaveButton: React.FC<SaveButtonProps> = ({
 
       layer.getSource()?.refresh();
 
+      setSaveCompleted(true);
+      setTimeout(() => {
+        setSaveCompleted(false);
+      }, 3000);
```
dnlkoch, line 136: Why do we need the `setTimeout`?

LukasLohoff: the success icon is removed after 3s. another option would be to keep it active until the form is modified again

dnlkoch: No, this is fine, thanks!

dnlkoch (approved):

LukasLohoff:
Thanks!

PR #978, opened 2023-06-30
Make loading of the fallback application config configurable

dnlkoch:
This introduces a new client config parameter `enableFallbackConfig` which determines if the client should load the fallback configuration if no application configuration (because not set explicitly or not loaded correctly) is available.

It also fixes an issue with empty locales as `i18n` might not have been initialized in the `catch` block.


ahennr (approved):

hwbllmnn (approved):

dnlkoch:
Thanks for the reviews!

PR #1029, opened 2023-07-21
Testsetup

FritzHoing:
Fixes the playwright config and adds aria-label as selectors.


e2e-tests/auth.setup.ts
```diff
@@ -1,17 +1,20 @@
...
-  + 'code&scope=openid&nonce=6cd0da9a-b45c-4a9e-931d-99b6725a148e');
+  await page.goto(`https://${process.env.HOST}/auth/realms/SHOGun`
+    + '/protocol/openid-connect/auth?client_id=shogun-client'
+    + `&redirect_uri=https%3A%2F%2F${process.env.HOST}%2Fclient%2F%3FapplicationId%3D21&state=9a983abe-3b0c-`
+    + '41cb-9b7e-1d9120956959&response_mode=fragment&response_type'
+    + '=code&scope=openid&nonce=72884466-0535-4a24-8c15-9e7f14d88a65');
 
+  // @ts-ignore
   await page.getByLabel('Username or email').fill(`${process.env.ADMIN_LOGIN}`);
+  // @ts-ignore
```
LukasLohoff: I think these might not be necessary, can you check if you still need the `@ts-ignore`?

src/components/Header/index.tsx
```diff
@@ -70,7 +70,9 @@ export const Header: React.FC<HeaderProps> = ({
     const items = [
       <SearchField
         key="search"
-        aria-label="search-bar"
+        aria-label='fritz'
```
LukasLohoff, line 73: was this intended?

src/e2e-tests/test-results/auth.setup.ts-authenticate-as-admin-setup-retry1/trace.zip
LukasLohoff, line 1: do we need the /test-result files or can we add them to `.gitignore`?

src/components/Header/index.tsx
FritzHoing: Good catch! It is adjusted now.

e2e-tests/auth.setup.ts
FritzHoing: You are right, i have removed the ts-ignore!

FritzHoing:
Thank you very much for the review @LukasLohoff !

playwright.config.js
```diff
@@ -17,29 +18,30 @@ export default defineConfig({
     open: 'never'
   }]],
   use: {
-    baseURL: process.env.HOST,
+    // das mal ausprobieren:
```
dnlkoch, line 21: This comment can be removed I assume?

src/e2e-tests/userMenu.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { test } from '@playwright/test';
+
+import { userMenu } from '@terrestris/shogun-e2e-tests/dist/shogun-gis-client/header/userMenu';
+
+test.use({
+  storageState: 'playwright/.auth/admin.json'
+});
+
+test('test', async ({
```
dnlkoch, line 9: We should apply some useful names to the test (here and in all others).

tsconfig.json
```diff
@@ -104,6 +102,9 @@
     "**.config.js"
   ],
   "include": [
-    "./src"
-  ]
+    "./src",
+    "e2e-tests/annotation.spec.ts",
+    "node_modules/@terrestris/shogun-e2e-tests/e2e-tests/shogun-gis-client/header/header",
```
dnlkoch, line 107: Is this needed?

tsconfig.json
```diff
@@ -104,6 +102,9 @@
...
+    "e2e-tests/annotation.spec.ts"
```
dnlkoch, line 108: Duplicate entry. (Is this needed anyway?)

PR #1039, opened 2023-07-26
Update tests

FritzHoing:
- Installs missing dependencies within Dockerfile
- Navigates to the desired feature for the imported tests to start

### DIFF
--- a/Dockerfile.e2e
+++ b/Dockerfile.e2e
@@ -1,13 +1,15 @@
-
 FROM mcr.microsoft.com/playwright:v1.30.0-focal@sha256:9e6eddb08b8566076ef778007859096bc90f8984236061c086eaede2fb60430b
 
 WORKDIR /app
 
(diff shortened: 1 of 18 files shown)


Dockerfile.e2e
```diff
@@ -1,13 +1,15 @@
...
 
 WORKDIR /app
 
 RUN npm install -D @playwright/test@v1.30.0
-RUN npm ci
 RUN npx playwright install --with-deps
 
+# Need to be installed since the package.json is not copied
+RUN npm i @terrestris/shogun-e2e-tests
+RUN npm i dotenv
```
dnlkoch, line 10: Should we pin fixed versions?

dnlkoch (approved):

Dockerfile.e2e
FritzHoing: Yes defenetly, versions are defined now!

PR #1164, opened 2023-09-14
Allow configurable (1:n) JSON fields in FeatureInfo and Edit forms

dnlkoch:
This adds support for creating a (configurable) UI for editing complex feature like objects that contain (1:n) JSON fields. Consider the following feature type holding the property `complexLikeProperty`:

```json
{
  "type": "Feature",
  "id": "my_feature_type.1909",
  "geometry": {
    "type": "MultiPolygon",
    "coordinates": […]
  },
  "geometry_name": "geom",
  "properties": {
    "complexLikeProperty": "[{\"id\" : 6, \"description\" : \"This is an example feature\", \"isPublic\" : true}]"
  },
  "bbox": […]
}
```

Backed by the following `editFormConfig`:

```json
{
  "title": "My complex feature type",
  "children": [{
    "propertyName": "complexLikeProperty",
    "displayName": "Complex Property",
    "component": "REFERENCE_TABLE",
    "tablePropertyName": "id",
    "editFormConfig": [{
      "propertyName": "id",
      "displayName": "ID",
      "component": "NUMBER",
      "readOnly": true
    }, {
      "propertyName": "description",
      "displayName": "Description",
      "component": "TEXTAREA",
      "required": true
    }, {
      "propertyName": "isPublic",
      "displayName": "Public",
      "component": "CHECKBOX"
    }]
  }]
} 
```

the following UI results:

Peek 2023-09-15 17-01

Requires [PR 384](https://github.com/terrestris/shogun-util/pull/384).


src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
```diff
@@ -159,6 +161,18 @@ export const EditFeatureFullForm: React.FC<EditFeatureFullFormProps> = ({
             properties[key] = [];
           }
         }
+
+        const isReference = tabConfigs[0].children?.find(cfg => {
```
LukasLohoff, line 165: since `isReference` is used only in an if-check, `Array.some()` would be sufficient here

LukasLohoff:
```suggestion
        const isReference = tabConfigs[0].children?.some(cfg => {
```

src/components/EditFeatureDrawer/EditReferenceTable/index.tsx
```diff
@@ -0,0 +1,379 @@
...
+import {
+  Button,
+  Modal,
+  Popconfirm,
+  Table
+} from 'antd';
+
+import {
+  FormInstance,
+  useForm,
```
LukasLohoff, line 23:
```suggestion
  useForm
```

src/components/EditFeatureDrawer/EditReferenceTable/index.tsx
```diff
@@ -0,0 +1,379 @@
...
+
+    if (!Array.isArray(value)) {
+      Logger.warn('Reference value must be of of type array.');
+      return;
+    }
+
+    const valueCopy = _cloneDeep(value);
+    valueCopy.forEach(v => {
+      Object.entries(v).forEach(([key, val]) => {
+        const isDate = formConfig?.some(cfg => cfg.propertyName === key && cfg.component === 'DATE');
```
LukasLohoff, line 176: might make sense to move these functions to a util so they can be reused

src/components/EditFeatureDrawer/EditReferenceTable/index.tsx
```diff
@@ -0,0 +1,379 @@
...
+
+  const onCancel = () => {
+    closeModal();
+  };
+
+  const onOk = async () => {
+    try {
+      await form.validateFields();
+    } catch (error) {
+      return;
```
LukasLohoff, line 318: is the silent return intentional? otherwise we should log something here

LukasLohoff (approved):
Very cool feature @dnlkoch !

src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
dnlkoch: You're absolutely right, fixed it.

src/components/EditFeatureDrawer/EditReferenceTable/index.tsx
dnlkoch: Added a logger.

dnlkoch:
Thanks for the detailed review!

PR #1210, opened 2023-09-28
Disable grouping in DisplayField

LukasLohoff:
See title.

### DIFF
--- a/src/components/DisplayField/index.tsx
+++ b/src/components/DisplayField/index.tsx
@@ -68,7 +68,9 @@ export const DisplayField: React.FC<DisplayFieldProps> = ({
   }
 
   if (Number.isFinite(value)) {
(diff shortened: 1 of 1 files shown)


mholthausen (approved):

src/components/DisplayField/index.tsx
```diff
@@ -68,7 +68,9 @@ export const DisplayField: React.FC<DisplayFieldProps> = ({
   }
 
   if (Number.isFinite(value)) {
-    displayText = new Intl.NumberFormat().format(Number(value));
+    displayText = new Intl.NumberFormat('de-DE', {
```
dnlkoch, line 71: Can we access the current application language here instead?

LukasLohoff: see https://github.com/terrestris/shogun-gis-client/pull/1213

PR #1214, opened 2023-09-29
Add layer attributions to print

LukasLohoff:


mholthausen:
Always adds the attributions of the layers to the parameters so that they are included in a printout

src/components/PrintForm/index.tsx
```diff
@@ -186,12 +238,16 @@ export const PrintForm: React.FC<PrintFormProps> = ({
       pManager.setDpi(pManager.getDpis()[0]);
       pManager.setLayout(pManager.getLayouts()[0]?.name);
 
+      dispatch(addCustomParam({
+        ['attributions']: getAttributions(pManager)
```
dnlkoch, line 242: Why do we need to pass the key `attributions` as array?

src/components/PrintForm/index.tsx
```diff
@@ -120,6 +132,46 @@ export const PrintForm: React.FC<PrintFormProps> = ({
...
+      return layerName &&
+        !(layerName.includes('react-geo')) &&
+        layer.getVisible() &&
+        !(layer instanceof OlLayerGroup) &&
+        layer !== printLayer;
+    }) as LayerType[];
+  }, [map]);
+
+  const getAttributions = useCallback((manager: MapFishPrintV3Manager) => {
+    if (!manager) {
```
dnlkoch, line 151: According to the typing this check shouldn't be needed, if this is actually optional, we should mark it accordingly.

dnlkoch (approved):

src/components/PrintForm/index.tsx
mholthausen: The unnecessary check got removed.

PR #1267, opened 2023-10-30
Uses always the first tab on new FI

mholthausen:
This MR update the `FeatureInfo` to use always the first layer tab when a new feature info is triggered

### DIFF
--- a/src/components/ToolMenu/FeatureInfo/index.tsx
+++ b/src/components/ToolMenu/FeatureInfo/index.tsx
@@ -68,6 +68,11 @@ export type FeatureInfoProps = FormProps & {
   enabled?: boolean;
 } & Partial<CoordinateInfoProps>;
 
+type MapLayer = {
+  layerName: string;
+  index: number;
+};
+
 export const FeatureInfo: React.FC<FeatureInfoProps> = ({
   enabled,
   ...restProps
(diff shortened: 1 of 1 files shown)


src/components/ToolMenu/FeatureInfo/index.tsx
```diff
@@ -260,6 +280,13 @@ export const FeatureInfo: React.FC<FeatureInfoProps> = ({
       serializedFeatures[layerName] = new OlFormatGeoJSON().writeFeatures(selectedFeatures);
     });
 
+    const layers: MapLayer[] = Object.keys(features).map(layerName => ({
+      layerName: layerName,
+      index: findMapLayerIndex(layerName)
+    })).sort((a, b) => b.index - a.index);
+
+    setActiveTabKey(layers[0].layerName);
```
dnlkoch, line 288: What happens if `layers[0]` is not available?

mholthausen: A check is added here to test if layers is present and has at least one value.

src/components/ToolMenu/FeatureInfo/index.tsx
```diff
@@ -122,6 +128,27 @@ export const FeatureInfo: React.FC<FeatureInfoProps> = ({
     return <></>;
   }
 
+  const changeActiveKey = (key: string) => {
+    setActiveTabKey(key);
+  };
+
+  const findMapLayerIndex = (layerName: string) => {
```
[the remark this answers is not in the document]
dnlkoch: Yes, you're right. Never mind.

dnlkoch (approved):

PR #1299, opened 2023-11-15
fix: fixing onChange by passing the correct state

FritzHoing:
Title Says it all. 

Two Prints in each format are appended.

[mapfish-print-report (6).pdf](https://github.com/terrestris/shogun-gis-client/files/13362912/mapfish-print-report.6.pdf)
[mapfish-print-report (5).pdf](https://github.com/terrestris/shogun-gis-client/files/13362914/mapfish-print-report.5.pdf)

### DIFF
--- a/src/components/PrintForm/LayoutSelect/index.tsx
+++ b/src/components/PrintForm/LayoutSelect/index.tsx
@@ -45,7 +45,7 @@ export const LayoutSelect: React.FC<LayoutSelectProps> = ({
     <Select
       placeholder={placeholder}
       value={layout}
-      onChange={setLayout}
+      onChange={layout}
       {...restProps}
     >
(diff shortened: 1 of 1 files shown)


src/components/PrintForm/LayoutSelect/index.tsx
```diff
@@ -45,7 +45,7 @@ export const LayoutSelect: React.FC<LayoutSelectProps> = ({
     <Select
       placeholder={placeholder}
       value={layout}
-      onChange={setLayout}
+      onChange={layout}
```
dnlkoch, line 48: `onChange` accepts a function, but `layout` is just a string.

dnlkoch (changes requested):

PR #1300, opened 2023-11-15
Draw upload: error handling and userinformation

FritzHoing:
On a successful upload a message will appear informing the user about a successful upload.
On an error, a error message will appear informing the user, that only .geojson files can be uploaded.

### DIFF
--- a/src/i18n/translations.ts
+++ b/src/i18n/translations.ts
@@ -45,6 +45,8 @@ export default {
         text: 'Anmerkung',
         modify: 'Bearbeitung',
         upload: 'Hochladen',
+        uploadSuccess: 'Die Datei wurde erfolgreich importiert',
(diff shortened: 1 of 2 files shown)


src/i18n/translations.ts
```diff
@@ -282,6 +284,8 @@ export default {
         text: 'Annotation',
         modify: 'Edit',
         upload: 'Upload',
+        uploadSuccess: 'The file was uploaded successfully',
+        uploadError: 'Only files of the type .geojson are supported',
```
dnlkoch, line 288: I assume the upload can also fail because of other reasons. Can we change it so something like "The import failed, please note that only .geojson files are supported".

FritzHoing: Good point! I have adjusted it according to your suggestion.

dnlkoch (approved):

PR #1301, opened 2023-11-15
feat: remove twitter from permalink

FritzHoing:
Removes twitter from permalink share temporally since its now renamed and a x-icon is not existing yet.

### DIFF
--- a/src/components/Permalink/index.tsx
+++ b/src/components/Permalink/index.tsx
@@ -5,7 +5,6 @@ import React, {
 import {
   CopyOutlined,
   MailOutlined,
-  TwitterOutlined,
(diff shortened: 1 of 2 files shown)


dnlkoch:
The X icon should be [available](https://fontawesome.com/icons/x-twitter?f=brands&s=solid), but I assume we don't use the latest version of font awesome yet.

dnlkoch:
> > The X icon should be [available](https://fontawesome.com/icons/x-twitter?f=brands&s=solid), but I assume we don't use the latest version of font awesome yet.
> 
> Yes that is correct!

Can we update then?

dnlkoch (approved):

PR #1309, opened 2023-11-17
fix: refactor loading animation

FritzHoing:
- Fixes the listener for image layer
- Change loading animation

loading-mask

### DIFF
--- a/src/components/AddLayerModal/index.tsx
+++ b/src/components/AddLayerModal/index.tsx
@@ -165,8 +165,6 @@ export const AddLayerModal: React.FC<AddLayerModalProps> = ({
         };
         layerToAdd.set('layerConfig', layerConfig);
         targetGroup.getLayers().push(layerToAdd);
-
-        map.getLayers().dispatchEvent('change:layers');
       }
     });
 
--- a/src/components/AddLayerModal/index.tsx
+++ b/src/components/AddLayerModal/index.tsx
@@ -165,6 +165,8 @@ export const AddLayerModal: React.FC<AddLayerModalProps> = ({
         };
         layerToAdd.set('layerConfig', layerConfig);
         targetGroup.getLayers().push(layerToAdd);
+
+        map.getLayers().dispatchEvent('change:layers');
       }
     });
 
--- a/src/components/ToolMenu/LayerTree/LoadingIndicator/index.less
+++ b/src/components/ToolMenu/LayerTree/LoadingIndicator/index.less
@@ -1,4 +1,4 @@
-.loading-animation {
+.layer-node-title-loading-animation {
   display: inline-block;
 
   span[class^="dot-"] {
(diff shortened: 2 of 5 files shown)


marcjansen:
Just an idea, is it possible to use `content` and pseudo-elements (like `::before`) to generate the `.` from the CSS? I think this might be better for usability.

FritzHoing:
> Just an idea, is it possible to use `content` and pseudo-elements (like `::before`) to generate the `.` from the CSS? I think this might be better for usability.

Good idea! I will try it out!

src/components/ToolMenu/LayerTree/LoadingIndicator/index.less
```diff
@@ -0,0 +1,74 @@
+.loading-mask {
```
dnlkoch, line 1: Not entirely sure, but this class name _might_ conflict with others, e.g. from antd (and it's not a mask in particular).

src/components/ToolMenu/LayerTree/LoadingIndicator/index.less
```diff
@@ -0,0 +1,74 @@
...
+ display: inline-block;
```
dnlkoch, line 2:
```suggestion
  display: inline-block;
```

src/components/ToolMenu/LayerTree/LoadingIndicator/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './index.less';
+
+export const LoadingIndicator: React.FC = (): JSX.Element => {
+  return (
+    <div className='loading-mask'>
```
dnlkoch, line 6:
```suggestion
    <div className="loading-mask">
```

src/components/ToolMenu/LayerTree/LoadingIndicator/index.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './index.less';
```
dnlkoch, line 2:
```suggestion

import './index.less';
```

src/components/AddLayerModal/index.tsx
```diff
@@ -152,6 +152,7 @@ export const AddLayerModal: React.FC<AddLayerModalProps> = ({
         };
         layerToAdd.set('layerConfig', layerConfig);
         targetGroup.getLayers().push(layerToAdd);
+        map.getLayers().dispatchEvent('change');
```
dnlkoch, line 155: Can't we reuse the existing `change:layers` event of the [root layer group](https://openlayers.org/en/latest/apidoc/module-ol_layer_Group-LayerGroup.html)?

FritzHoing: Yes that is correct! It will be adjusted

src/components/ToolMenu/LayerTree/LoadingIndicator/index.less
FritzHoing: i have changed it to "loading-animation"

dnlkoch: That's still quite generic, probably something like `layer-node-title-loading-animation`?

dnlkoch (approved):

PR #1328, opened 2023-11-22
Tool menu layout

FritzHoing:
This MR changes the color of the active Button within the toolbar.

Instead of a default Black the secondary color is now used, which can be adjusted dynamically via the admin-client.

styling

### DIFF
--- a/src/components/ToolMenu/Draw/index.less
+++ b/src/components/ToolMenu/Draw/index.less
@@ -6,7 +6,7 @@
 
     &.btn-pressed {
       background-color: white;
-      color: var(--ant-primary-10);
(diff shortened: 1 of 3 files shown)


dnlkoch (approved):
Generally I'm totally fine with this. Just wondering, have you checked to just set the button type to `primary`?

FritzHoing:
> Generally I'm totally fine with this. Just wondering, have you checked to just set the button type to `primary`?

Yes i did, but then the background color of the buttons would be very dominant and would not fit to the rest of the drawer. Of course changing the background color would be an option but i felt that this PR might already do it's purpose.

PR #1399, opened 2024-01-10
Adds search result zoom offset

mholthausen:
This PR uses a new feature of react-geo's `SearchResultPanel` and overwrites the `onClick` property. A function is used to move the zoom center on a feature to the right by the actual width of the tool-menu.

In addition, the `NominatimSearch` component, which is irrelevant for this feature, was adapted to [other changes](https://github.com/terrestris/react-geo/commit/f1e0df793575805e57d7b548c3aaca844147e2bd) from react-geo.

### DIFF
--- a/src/components/MultiSearch/index.tsx
+++ b/src/components/MultiSearch/index.tsx
@@ -108,6 +108,10 @@ export type SolrQueryConfig = {
   'hl.requireFieldMatch'?: boolean;
 };
 
+export type Item = {
+  feature: OlFeature;
+};
+
 export const MultiSearch: React.FC<MultiSearchProps> = ({
   useNominatim
 }): JSX.Element => {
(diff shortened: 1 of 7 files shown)


src/global.d.ts
```diff
@@ -33,6 +33,7 @@ declare module 'clientConfig' {
     minChars?: number;
     coreName?: string;
     solrQueryConfig?: SolrQueryConfig;
+    featureResultZoomOffset?: number;
```
dnlkoch, line 36: Do we need to have this configurable? I would favour to have this detected automatically (if possible).

src/components/MultiSearch/index.tsx
```diff
@@ -545,6 +547,34 @@ export const MultiSearch: React.FC<MultiSearchProps> = ({
         searchTerms={searchValue.split(' ')}
         actionsCreator={actionsCreator}
         layerStyle={layerStyle}
+        onClick={(item: any) => {
```
dnlkoch, line 550: Is there a type for `item` available? If not, can we create one?

src/global.d.ts
mholthausen: Yes, that works too. Some other changes have been reversed and the files have been adjusted accordingly in [50832de](https://github.com/terrestris/shogun-gis-client/pull/1399/commits/50832de8f122b0301791808cfcb479659376d980).

src/components/MultiSearch/index.tsx
mholthausen: No, there wasn't a type for `item` yet. It was also added in [50832de](https://github.com/terrestris/shogun-gis-client/pull/1399/commits/50832de8f122b0301791808cfcb479659376d980) and is now used there.

dnlkoch (approved):

PR #1430, opened 2024-01-22
fix: prevents the application from crashing when a corrupt file is uploaded

faouziH21:
This fixes a bug where when uploading a corrupt (Geo)JSON file the website would crash. Additionally added a new function that checks the viability of the uploaded file. This should enhance the readability of the code.

### DIFF
--- a/src/components/ToolMenu/Draw/index.tsx
+++ b/src/components/ToolMenu/Draw/index.tsx
@@ -131,6 +131,10 @@ export const Draw: React.FC<DrawProps> = ({
     const uploadedFiles = e.target.files;
     if (checkValidityOfUploadFile(e)){
       onGeoJSONUpload(uploadedFiles![0]);
+    } else {
+
+      message.error(t('Draw.uploadError'));
+
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/components/ToolMenu/Draw/index.tsx
```diff
@@ -110,8 +110,9 @@ export const Draw: React.FC<DrawProps> = ({
     }
   };
 
-  const onUploadChange = (e: ChangeEvent<HTMLInputElement>) => {
+  const checkValidityOfUploadFile = (e: ChangeEvent<HTMLInputElement>) => {
     const uploadedFiles = e.target.files;
+    let validitiy = false;
```
FritzHoing, line 115: Is it maybe better to make use of a state here?

jansule (commented):
It looks like there will be no feedback if the validity is false, as the only negative feedback is in the catch block

src/components/ToolMenu/Draw/index.tsx
faouziH21: Not sure on how it would be implemented. This keeps the function simple and readable.

faouziH21: No this method is used in a condition and should not have any side effects.

faouziH21:
> It looks like there will be no feedback if the validity is false, as the only negative feedback is in the catch block

fixed in [e6ab02a](https://github.com/terrestris/shogun-gis-client/pull/1430/commits/e6ab02af598add511898dc219ea055cf9793c9b3)

dnlkoch (approved):

PR #1515, opened 2024-03-14
Update dependencies for ol and react

mholthausen:
This MR is intended to use the new [mapfish-print-manager](https://github.com/terrestris/mapfish-print-manager/pull/468), [ol-util](https://github.com/terrestris/ol-util/pull/1359), [react-util](https://github.com/terrestris/react-util/pull/627) and [react-geo](https://github.com/terrestris/react-geo/pull/3773) (and finally https://github.com/terrestris/react-geo/tree/rg-future) versions. After these have been released, the updated `package-lock.json` must be added.

## TODO
* Toolbar component was removed in https://github.com/terrestris/react-geo/pull/3503, see also https://github.com/terrestris/react-geo/discussions/3494, so it has to be determined how it should be replaced in `EditFeatureDrawer/EditFeatureGeometryToolbar`
*


simonseyock (approved):
I can't really see if the antd and react updates work that easily. I hope the tests still work in theory?

ahennr:
Since this is a major update of OpenLayers, antd and further utils I prefer to declare this as a breaking change. Could you please change the commit message accordingly @mholthausen ?

mholthausen:
Yes, thanks. Updated the description and the commit. Also added some adjustments to the tests and components as they didn't run properly at first.
There more work to do, especially to to fulfill the typecheck.

dnlkoch:
Note: We tried to switch to vite as bundler since the failing start of the application while using `ol` as shared module seems to be an open issue in [webpack](https://github.com/webpack/webpack/issues/16125).

hwbllmnn:
I've added an override for ol to fix the remaining module federation warnings, dependencies have been updated and sorted again. The example plugin can be successfully loaded and the client seems to be working fine after a quick test.

PR #1538, opened 2024-04-24
fix: pin versions of ol and ol-mapbox-style

ahennr:
and remove openlayers from shared modules in module federation config.


dnlkoch:
Replaced by #1515.

PR #1552, opened 2024-04-30
Fix several code smells that a SonarQube check identified

marcjansen:
This PR fixes 3 types of SonarQube detected code smells:

* No duplicated key in the Jest config => dfd44be857b75c4ec78beb7244ec9972b8efcc21
* using `[].forEach` and not `[].map` in one instance => 59f0d60c3b101874c9884d263331404561dde8dd
* not using type intersections like `export type HumptyProps = {} | SomeOtherProps` or `export type HumptyProps = {} | Partial<SomeOtherProps>` => c77d18325bec2bcec0b2d7390bde7db19d5f0875

WRT the first two commits above I am quite quite sure they do the correct thing™

I am not so sure about the effects of c77d18325bec2bcec0b2d7390bde7db19d5f0875. @dnlkoch nodded an OK after a quick 15 seconds explanation around the coffee machine ;-)

Please review.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,7 +5,6 @@ module.exports = {
     KEYCLOAK_REALM: 'SHOGun',
     KEYCLOAK_CLIENT_ID: 'shogun-client'
   },
-  moduleFileExtensions: ['ts', 'js'],
   transform: {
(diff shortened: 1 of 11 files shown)


LukasLohoff (approved):

marcjansen:
Thanks @LukasLohoff

PR #1555, opened 2024-04-30
fix: adjust id name to match database

TreffN:
The id name changes when working with the database. This PR adjusts this.

Please review @dnlkoch

### DIFF
--- a/src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
+++ b/src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
@@ -93,7 +93,7 @@ export const EditFeatureFullForm: React.FC<EditFeatureFullFormProps> = ({
 
     const updatedFeatures = await executeGetFeature({
       layer: layer,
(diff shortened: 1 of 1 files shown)


src/components/EditFeatureDrawer/EditFeatureFullForm/index.tsx
```diff
@@ -93,7 +93,7 @@ export const EditFeatureFullForm: React.FC<EditFeatureFullFormProps> = ({
 
     const updatedFeatures = await executeGetFeature({
       layer: layer,
-      filter: equalTo('id', id)
+      filter: equalTo('ogc_fid', id)
```
dnlkoch, line 96: Actually both values aren't quite sophisticated since the id column can be any numeric field of the feature type. We should think about making it configurable or getting it dynamically from the `DescribeFeatureType`.

TreffN:
This branch is closed, because this hardcoded version can be replaced with a dynamic version.

PR #1560, opened 2024-05-02
feat: make loading mask configurable

FritzHoing:
This MR uses the logo defined under 'logoPath' within the client-config as a loading image.

It is also saved within the cookies to ensure a faster image loading process.

co authored by @AmandaTamanda & Noémie

### DIFF
--- a/resources/public/index.ejs
+++ b/resources/public/index.ejs
@@ -13,7 +13,7 @@
     <base href="<%= htmlWebpackPlugin.options.appPrefix %>">
     <link rel="shortcut icon" type="image/x-icon" href="favicon.ico">
     <div id="loadmask" class="loadmask">
-      <img src="./shogun_spinner.gif" alt="Loading, please wait…">
+      <img id="loadmaskImage" alt="Loading, please wait…">
     </div>
     <title><%= htmlWebpackPlugin.options.title %></title>
     <style>
@@ -77,6 +77,18 @@
       if (!window.clientConfig) {
         console.warn('No config found, client defaults will be used.');
         window.clientConfig = {};
+      }
+    </script>
+    <script>
+      const regex = /\?applicationId=(\d+)/;
+      const appId = location.search.match(regex)[1]
+      let logoPath = './shogun_spinner.gif';
+      if (appId) {
+        logoPath = localStorage.getItem(`SHOGun_Logo_Path_${appId}`);
+      }
+      const loadingImageElement = document.getElementById('loadmaskImage');
+      if (logoPath && loadingImageElement) {
+        loadingImageElement.src = logoPath;
       }
     </script>
     <div id="app" class="app"></div>
--- a/resources/public/index.ejs
+++ b/resources/public/index.ejs
@@ -80,8 +80,8 @@
       }
     </script>
     <script>
-      const regex = /\?applicationId=(\d+)/;
-      const appId = location.search.match(regex)[1]
+      const regex = /\applicationId=(\d+)/;
+      const appId = location.search.match(regex)[1];
       let logoPath = './shogun_spinner.gif';
       if (appId) {
         logoPath = localStorage.getItem(`SHOGun_Logo_Path_${appId}`);
--- a/resources/public/index.ejs
+++ b/resources/public/index.ejs
@@ -13,7 +13,7 @@
     <base href="<%= htmlWebpackPlugin.options.appPrefix %>">
     <link rel="shortcut icon" type="image/x-icon" href="favicon.ico">
     <div id="loadmask" class="loadmask">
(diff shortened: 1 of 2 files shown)


resources/public/index.ejs
```diff
@@ -13,7 +13,7 @@
     <base href="<%= htmlWebpackPlugin.options.appPrefix %>">
     <link rel="shortcut icon" type="image/x-icon" href="favicon.ico">
     <div id="loadmask" class="loadmask">
-      <img src="./shogun_spinner.gif" alt="Loading, please wait…">
+      <img id="loadmaskImage" alt="Loading, please wait…">
```
KaiVolland, line 16:
```suggestion
      <img id="loadmask-image" alt="Loading, please wait…">
```

its more common to use kebap case in HTML and css

resources/public/index.ejs
```diff
@@ -79,6 +79,18 @@
...
       }
     </script>
+    <script>
+      const regex = /\?applicationId=(\d+)/;
+      const appId = location.search.match(regex)[1]
+      let logoPath = './shogun_spinner.gif';
+      if (appId) {
+        logoPath = localStorage.getItem(`SHOGun_Logo_Path_${appId}`);
+      }
+      const loadingImageElement = document.getElementById('loadmaskImage');
```
KaiVolland, line 89:
```suggestion
      const loadingImageElement = document.getElementById('loadmask-image');
```
see above

src/bootstrap.tsx
```diff
@@ -287,6 +287,17 @@ const setUserToStore = async (user?: User) => {
   store.dispatch(setUser(user));
 };
 
+const setLoadingImage = (path = './shogun_spinner.gif', id: number) => {
+  const loadingImageElement = document.getElementById('loadmaskImage') as HTMLImageElement;
```
KaiVolland, line 291:
```suggestion
  const loadingImageElement = document.getElementById('loadmask-image') as HTMLImageElement;
```

KaiVolland (approved):
Just one minor note regarding naming conventions in HTML/CSS.

resources/public/index.ejs
```diff
@@ -79,6 +79,18 @@
         window.clientConfig = {};
       }
     </script>
+    <script>
+      const regex = /\?applicationId=(\d+)/;
+      const appId = location.search.match(regex)[1]
```
dnlkoch, line 84:
```suggestion
      const appId = location.search.match(regex)[1];
```

resources/public/index.ejs
```diff
@@ -79,6 +79,18 @@
         window.clientConfig = {};
       }
     </script>
+    <script>
+      const regex = /\?applicationId=(\d+)/;
```
dnlkoch, line 83: Actually this requires that the `applicationId` is the first query parameter. We should probably change it to become more flexible with `/\applicationId=(\d+)/` (untested!).

src/bootstrap.tsx
```diff
@@ -287,6 +287,17 @@ const setUserToStore = async (user?: User) => {
   store.dispatch(setUser(user));
 };
 
+const setLoadingImage = (path = './shogun_spinner.gif', id: number) => {
```
dnlkoch, line 290: Non blocking, but it's more convenient to place the optional arguments at the end of the signature:

```suggestion
const setLoadingImage = (id: number, path = './shogun_spinner.gif') => {
```

src/bootstrap.tsx
```diff
@@ -287,6 +287,17 @@ const setUserToStore = async (user?: User) => {
...
+  const loadingImageElement = document.getElementById('loadmask-image') as HTMLImageElement;
+  if (!loadingImageElement) {
+    return;
+  }
+  loadingImageElement.src = path;
+  if (id) {
```
dnlkoch, line 296: According to the signature this can't be falsy.

src/bootstrap.tsx
```diff
@@ -650,10 +661,10 @@ const renderApp = async () => {
     if (!applicationId && !ClientConfiguration.enableFallbackConfig && !ClientConfiguration.staticAppConfigUrl) {
       throw new Error(LoadingErrorCode.APP_ID_NOT_SET);
     }
-
     let appConfig;
     if (applicationId && client) {
       appConfig = await getApplicationConfiguration(client, applicationId);
+      setLoadingImage(appConfig?.clientConfig?.theme?.logoPath, applicationId);
```
dnlkoch, line 667: See above.

dnlkoch (commented):
That's a very elegant approach, thanks!

resources/public/index.ejs
FritzHoing: Thank you! I have tested it and it works as expected

dnlkoch (approved):

PR #1565, opened 2024-05-08
test: change sonarqube configuration to cause job to fail when quality gate fails

sdobbert:
This sets the sonar.qualitygate.wait=true parameter which forces the sonarqube step to poll the SonarQube instance until the quality gate status is available. 
It causes the sonarqube analysis step to fail any time the quality gate fails.

Also increases the pipeline duration!

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -2,3 +2,6 @@ sonar.projectBaseDir=/home/runner/work/shogun-demo-client/shogun-demo-client
 sonar.testExecutionReportPaths=./coverage/jest-report.xml
 sonar.tests=src
 sonar.test.inclusions=**/*.spec.tsx,**/*.spec.ts
+
+# quality gate rules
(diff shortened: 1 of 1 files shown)


dnlkoch (commented):
Not entirely sure if I understand it correctly, but this might have unexpected results:

- During the [pull request pipeline](https://github.com/terrestris/shogun-gis-client/blob/main/.github/workflows/on-pull-request.yml) no sonarqube check is performed.
- After being merged, the check is being performed on the `main` branch.
- If the check _would_ fail, no build of the `main` would happen.
- In this case no image would be created and the `gh-pages` wouldn't be updated.

This seems quite confusing to me since I would expect a succeeding main branch pipeline after merging a succeeding pull request pipeline.

dnlkoch:
Closing this one, please reopen @sdobbert if needed.

PR #1570, opened 2024-05-10
Fix bumpy loading indicator

faouziH21:
This PR removes previous loading dots animation and adds a turning loading animation under the layer transparency slider. This is done to avoid bumps while loading. If a name was long enough a new line would be added to make place for the loading dots, which caused bumpy UI. This PR attempts to remove this bug permanently by making a permanent and specified place for the loading icon without using too much space.  

PR look: 
image

Previous look: 
image


FritzHoing:
Nice @faouziH21! But maybe the loading symbol is using too much space, if many layers are used?

KaiVolland:
Im not sure. :thinking: Dienst this add just another space below? What about replacing the menu item or checkebox while playing?

faouziH21:
> Nice @faouziH21! But maybe the loading symbol is using too much space, if many layers are used?

I am not sure... tested it with 5 Layers... it looks pretty normal for me...
image

no loading indicators: 
image

faouziH21:
> Im not sure. 🤔 Dienst this add just another space below? What about replacing the menu icon or checkbox while playing?

Good consideration yet if the layer takes a lot of time to load and the user need to access or uncheck the layer (details) they would have to wait. Which seems pretty counter intuitive. Of-course this might not always be the case but with multiple layers the layers do in fact take some time to load. That is one of the reasons why I decided to take up a little more space than to "replace" a feature.

faouziH21:
changes regarding this issue have been moved to [here](https://github.com/terrestris/shogun-gis-client/pull/1603)

PR #1582, opened 2024-05-15
feat: creates ability to select app specific print templates

AmandaTamanda:
Gives the possibility to add several templates for the print tool. Each app therefore has the possibility of having a combination different templates.

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -688,9 +688,9 @@ const renderApp = async () => {
       i18n.changeLanguage(defaultLanguage);
     }
 
-    const printAppConf = appConfig?.clientConfig?.printApp;
-    if (printAppConf) {
-      store.dispatch(setPrintApp(printAppConf));
+    const printApp = appConfig?.clientConfig?.printApp;
+    if (printApp) {
+      store.dispatch(setPrintApp(printApp));
     }
 
     const style = parseTheme(appConfig?.clientConfig?.theme);
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -97,6 +97,7 @@ import {
 import {
   setLogoPath
 } from './store/logoPath';
+import { setPrintApp } from './store/print';
 import {
   setSearchEngines
 } from './store/searchEngines';
@@ -115,7 +116,6 @@ import {
 } from './store/user';
 
 import './index.less';
-import { setPrintApp } from './store/print';
 
 // TODO: extend antd properties too
 export interface ThemeProperties extends React.CSSProperties {
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -688,10 +688,10 @@ const renderApp = async () => {
       i18n.changeLanguage(defaultLanguage);
     }
 
-    const printAppTemplates
-      = appConfig?.clientConfig?.printApp;
-
-    store.dispatch(setPrintApp(printAppTemplates));
+    const printAppConf = appConfig?.clientConfig?.printApp;
+    if (printAppConf) {
+      store.dispatch(setPrintApp(printAppConf));
+    }
 
     const style = parseTheme(appConfig?.clientConfig?.theme);
 
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -688,9 +688,10 @@ const renderApp = async () => {
       i18n.changeLanguage(defaultLanguage);
     }
 
-    const printApp = appConfig?.clientConfig?.printApp;
+    const printAppTemplates
+     = appConfig?.clientConfig?.printApp;
 
-    store.dispatch(setPrintApp(printApp));
+    store.dispatch(setPrintApp(printAppTemplates));
 
     const style = parseTheme(appConfig?.clientConfig?.theme);
 
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -115,6 +115,7 @@ import {
 } from './store/user';
 
 import './index.less';
+import { setPrintApp } from './store/print';
 
 // TODO: extend antd properties too
 export interface ThemeProperties extends React.CSSProperties {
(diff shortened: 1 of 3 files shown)


src/bootstrap.tsx
```diff
@@ -687,6 +688,10 @@ const renderApp = async () => {
       i18n.changeLanguage(defaultLanguage);
     }
 
+    const printApp = appConfig?.clientConfig?.printApp;
+
+    store.dispatch(setPrintApp(printApp));
```
dnlkoch, line 693: Since `printApp` is optional, what happens if it's not defined?

src/bootstrap.tsx
```diff
@@ -687,6 +688,10 @@ const renderApp = async () => {
       i18n.changeLanguage(defaultLanguage);
     }
 
+    const printApp = appConfig?.clientConfig?.printApp;
```
dnlkoch, line 691: Not entirely sure if needed, but do we need to add the new field to the [shogun-util](https://github.com/terrestris/shogun-util/blob/main/src/model/Application.ts#L53)?

src/components/PrintForm/index.tsx
```diff
@@ -225,7 +226,9 @@ export const PrintForm: React.FC<PrintFormProps> = ({
       // Use locale print app if available.
       // Implies that a print app with the language code exists.
       const apps = pManager.getPrintApps();
-
+      if (printApp){
```
dnlkoch, line 229:
```suggestion
      if (printApp) {
```

src/components/PrintForm/index.tsx
```diff
@@ -225,7 +226,9 @@ export const PrintForm: React.FC<PrintFormProps> = ({
...
+        await pManager.setPrintApp(printApp);
+      }
       if (apps && currentLanguageCode && apps.includes(currentLanguageCode)) {
         await pManager.setPrintApp(currentLanguageCode);
```
dnlkoch, line 233: This would potentially override the custom printApp. Not sure if this intended?

dnlkoch (commented):
Very nice @AmandaTamanda! That's a very useful addition.

src/bootstrap.tsx
AmandaTamanda: So far, when not defined in defaults to the defult templates setting, i tested it several times to make sure

AmandaTamanda:
> Very nice @AmandaTamanda! That's a very useful addition.

Thank you for the review :)

src/bootstrap.tsx
AmandaTamanda: > Not entirely sure if needed, but do we need to add the new field to the [shogun-util](https://github.com/terrestris/shogun-util/blob/main/src/model/Application.ts#L53)?

The client app doesn't start without it, but maybe there is something else behind that? I am open to suggestions.

src/bootstrap.tsx
```diff
@@ -687,6 +688,11 @@ const renderApp = async () => {
       i18n.changeLanguage(defaultLanguage);
     }
 
+    const printAppTemplates
```
dnlkoch, line 691: The name of the variable is misleading, it's a single name of a print app and not a set of templates.

src/bootstrap.tsx
```diff
@@ -687,6 +688,11 @@ const renderApp = async () => {
...
+      = appConfig?.clientConfig?.printApp;
+
+    store.dispatch(setPrintApp(printAppTemplates));
```
dnlkoch, line 694: Please wrap the dispatch in a condition. So it will be applies to the store if set in the configuration only.

src/components/PrintForm/index.tsx
```diff
@@ -93,6 +93,7 @@ export const PrintForm: React.FC<PrintFormProps> = ({
 
   const customMapParams = useAppSelector(state => state.print.customMapParams);
   const customParams = useAppSelector(state => state.print.customParams);
+  const printAppTemplates = useAppSelector(state => state.print.printApp);
```
dnlkoch, line 96: See comment regarding the variable name from above.

src/bootstrap.tsx
dnlkoch: See comment in new review.

src/components/PrintForm/index.tsx
dnlkoch: This issue still seems to exist. Do you see any chance to fix it?

AmandaTamanda: How do I fix that?

> This issue still seems to exist. Do you see any chance to fix it?

dnlkoch: The issue is as follows:

- the mapfish print servlet [may serve different print apps](https://mapfish.github.io/mapfish-print-doc/api.html#apps).
- line 232 checks if there is any print app available that contains the current language code (e.g. `de`).
- if so, this print app will be used.

Actually this will override the print app set by the configuration (lines 229 - 231). We should check if this behaviour could be optimized.

AmandaTamanda: Ok, thanks for the clarification. I have made the changes

src/components/PrintForm/index.tsx
```diff
@@ -225,9 +226,12 @@ export const PrintForm: React.FC<PrintFormProps> = ({
...
       const apps = pManager.getPrintApps();
-
-      if (apps && currentLanguageCode && apps.includes(currentLanguageCode)) {
-        await pManager.setPrintApp(currentLanguageCode);
+      if (!printApp) {
+        if (apps && currentLanguageCode && apps.includes(currentLanguageCode)) {
+          await pManager.setPrintApp(currentLanguageCode);
+        }
+      } else {
+          await pManager.setPrintApp(printApp);
```
dnlkoch, line 234:
```suggestion
        await pManager.setPrintApp(printApp);
```

dnlkoch (approved):

PR #1620, opened 2024-05-28
Hide user chip in public applications

LukasLohoff:
Title says it all.


KaiVolland (approved):

LukasLohoff:
Thanks!

KaiVolland:
What about applications that are public but you still can login. For e.g. you have edit permission only when you are logged in?

PR #1626, opened 2024-05-31
Create anchor elements for linklike values in DisplayField

KaiVolland:
This creates anchor elements (`<a href="" />`) elements for values that seem to be links.

### DIFF
--- a/src/components/DisplayField/index.tsx
+++ b/src/components/DisplayField/index.tsx
@@ -6,6 +6,7 @@ import {
   UploadFile
 } from 'antd';
 
+import isString from 'lodash/isString';
 import {
   isMoment
 } from 'moment';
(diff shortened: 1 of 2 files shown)


src/components/DisplayField/index.tsx
```diff
@@ -68,17 +69,45 @@ export const DisplayField: React.FC<DisplayFieldProps> = ({
...
+  }
+
+  const isUrl = (candidate: string) => {
+    const lowerCandidate = candidate.toLowerCase();
+    const protocols = ['http://', 'https://', 'ftp://', 'file://', 'mailto:'];
+    if (protocols.some(protocol => lowerCandidate.startsWith(protocol))) {
+      return true;
+    }
+    // windows (network) paths
+    if (candidate.startsWith('\\\\')) {
```
FritzHoing, line 92: shouldn't this be '////' ?

KaiVolland: Windows uses backslashes. But there should be only two not four. I fixed that in 959d914

FritzHoing (approved):

PR #1651, opened 2024-06-17
Redirect to login after logout

dnlkoch:
This redirects to the login page after logging out instead of reloading the client which may case an unexpected "You're not allowed to see this application" error.


dnlkoch:
Thanks for the review!

PR #1671, opened 2024-06-27
feat: make upload options configurable

TreffN:
This PR makes the uploading options in the layer tree configurable by setting:
    "name": "tree",
    "config": {
      "visible": true,
      "uploadTools": [
        "addWMS",
        "dataUpload"
      ]
    }
in the tool config in shogun-admin as described in [#22030](https://redmine.intranet.terrestris.de/issues/22030?include=journals&journals=all)

Different upload options are visible depending on the tool-configuration in shogun-admin:
Konfigurierbarkeit1
Konfigurierbarkeit2

Please review :)

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -91,6 +91,7 @@ import {
   EditLevel
 } from './store/editFeature';
 import { setFeatureInfoActiveCopyTools } from './store/featureInfo';
+import { setLayerTreeActiveUploadTools } from './store/layerTree';
 import {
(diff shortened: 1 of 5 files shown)


ahennr:
Generally :+1:, however I would prefer to change the commit type to `feat` since this is not a fix (IMHO). But, you can leave this as is as well, you decide.

ahennr (approved):

LukasLohoff (approved):
Looks great! Thanks

PR #1720, opened 2024-08-01
feat: show legends per default

TreffN:
This PR adds an option to set the layer legends visible by default when the layer is visible, so you don't need to select the legend in the context menu. 
Settings in the tool-config in SHOGun-Admin:
{
    "name": "tree",
    "config": {
      "showLegends": false,
      "visible": true
    }
  },
Legenden1

changing the showLegends config to true gives this initial view in the SHOGun-Gis-Client:
Legenden2

Please review :)

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -91,7 +91,10 @@ import {
   EditLevel
 } from './store/editFeature';
 import { setFeatureInfoActiveCopyTools } from './store/featureInfo';
-import { setLayerTreeActiveUploadTools } from './store/layerTree';
+import { 
+  setLayerTreeActiveUploadTools,
+  setLayerTreeShowLegends
+} from './store/layerTree';
 import {
   setLegal
 } from './store/legal';
@@ -270,6 +273,9 @@ const setApplicationToStore = async (application?: Application) => {
         }
         if (tool.name === 'tree' && Array.isArray(tool.config.uploadTools)) {
           store.dispatch(setLayerTreeActiveUploadTools(tool.config.uploadTools));
+        }
+        if (tool.name === 'tree' && tool.config.showLegends) {
+          store.dispatch(setLayerTreeShowLegends(tool.config.showLegends));
         }
       });
     store.dispatch(setAvailableTools(availableTools));
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -91,7 +91,7 @@ import {
   EditLevel
 } from './store/editFeature';
 import { setFeatureInfoActiveCopyTools } from './store/featureInfo';
-import { 
+import {
   setLayerTreeActiveUploadTools,
   setLayerTreeShowLegends
 } from './store/layerTree';
--- a/src/components/ToolMenu/LayerTree/index.tsx
+++ b/src/components/ToolMenu/LayerTree/index.tsx
@@ -40,7 +40,6 @@ import {
 
 import useAppSelector from '../../../hooks/useAppSelector';
 import useSHOGunAPIClient from '../../../hooks/useSHOGunAPIClient';
-import { UploadTools } from '../../../store/layerTree';
 
 import WmsTimeSlider from '../../WmsTimeSlider';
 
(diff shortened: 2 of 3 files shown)


src/bootstrap.tsx
```diff
@@ -91,7 +91,10 @@ import {
   EditLevel
 } from './store/editFeature';
 import { setFeatureInfoActiveCopyTools } from './store/featureInfo';
-import { setLayerTreeActiveUploadTools } from './store/layerTree';
+import {
```
faouziH21, line 94:
``` suggestion 
import {
```

src/components/ToolMenu/LayerTree/index.tsx
```diff
@@ -72,11 +74,15 @@ export const LayerTree: React.FC<LayerTreeProps> = ({
     t
   } = useTranslation();
 
-  const [visibleLegendsIds, setVisibleLegendsIds] = useState<string[]>([]);
-  const [layerTileLoadCounter, setLayerTileLoadCounter] = useState<LayerTileLoadCounter>({});
+  const showLegendsState = useAppSelector(state => state.layerTree.showLegends);
+  //check if showLegends is defined
```
faouziH21, line 78:
```suggestion 
// check if showLegends is defined
```

faouziH21:
Nice! I Have one question: Does this cause a lot of scrolling in the case that there are a lot of layers? If so maybe we should consider making the legend collapsible. But all in all well done :D

src/bootstrap.tsx
TreffN: fixed in new commit

src/components/ToolMenu/LayerTree/index.tsx
TreffN: fixed in newest commit

LukasLohoff (approved):

src/components/ToolMenu/LayerTree/index.tsx
```diff
@@ -61,6 +62,7 @@ export type LayerTileLoadCounter = {
 export type LayerTreeConfig = {
   enabled?: boolean;
   activeUploadTools?: UploadTools[];
+  showLegends?: boolean;
 };
```
dnlkoch, line 66: Not directly related to the changes in this PR, but it looks like the type `LayerTreeConfig` is used in the store only. Probably we should think about moving it over.

src/components/ToolMenu/LayerTree/index.tsx
```diff
@@ -72,11 +74,15 @@ export const LayerTree: React.FC<LayerTreeProps> = ({
     t
   } = useTranslation();
 
-  const [visibleLegendsIds, setVisibleLegendsIds] = useState<string[]>([]);
-  const [layerTileLoadCounter, setLayerTileLoadCounter] = useState<LayerTileLoadCounter>({});
+  const showLegendsState = useAppSelector(state => state.layerTree.showLegends);
+  // check if showLegends is defined
+  const showLegends: boolean = showLegendsState ? showLegendsState : false;
```
dnlkoch, line 79: Is this really needed? Can't we use `showLegendsState` directly?

src/components/ToolMenu/LayerTree/index.tsx
```diff
@@ -72,11 +74,15 @@ export const LayerTree: React.FC<LayerTreeProps> = ({
...
 
   const initialLayersUid = map?.getAllLayers().map(l => getUid(l));
 
+  const [visibleLegendsIds, setVisibleLegendsIds] = useState<string[]> (showLegends? initialLayersUid ?? [] : []);
```
dnlkoch, line 83:
```suggestion
  const [visibleLegendsIds, setVisibleLegendsIds] = useState<string[]> (showLegends ? initialLayersUid ?? [] : []);
```

dnlkoch (approved):

src/components/ToolMenu/LayerTree/index.tsx
TreffN: Moved the type LayerTreeConfig to the store

src/components/ToolMenu/LayerTree/index.tsx
TreffN: Only use of showLegendsState in the new commit

src/components/ToolMenu/LayerTree/index.tsx
TreffN: see new commit

LukasLohoff (approved):

PR #1726, opened 2024-08-06
Set layer visibility on search result click

LukasLohoff:
- sets layer visibility on search result click
- also sets visibility of parent layer groups
- collapse search result panel


KaiVolland (approved):

LukasLohoff:
Thanks!

PR #1743, opened 2024-08-26
feat: Introduce docs button

faouziH21:
prev:
image
now: 
image

### DIFF
--- a/src/components/DocumentationButton/index.tsx
+++ b/src/components/DocumentationButton/index.tsx
@@ -1,0 +1,36 @@
+
+import React from 'react';
+
+import { faCircleQuestion } from '@fortawesome/free-solid-svg-icons';
+import {
+  FontAwesomeIcon
+} from '@fortawesome/react-fontawesome';
+
+import SimpleButton, { SimpleButtonProps } from '@terrestris/react-geo/dist/Button/SimpleButton/SimpleButton';
+
+export type DocumentationButtonProps = SimpleButtonProps;
+
+const defaultClassName = 'documentationbutton';
+export const DocumentationButton: React.FC<DocumentationButtonProps> = ({
+  className
+}) => {
+
+  const finalClassName = className
+    ? `${defaultClassName} ${className}`
+    : defaultClassName;
+
+  return (
+    <SimpleButton
+      onClick={() => window.open('/gis-docs', '_blank')}
+      className={finalClassName}
+      icon={
+        <FontAwesomeIcon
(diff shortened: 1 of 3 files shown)


src/components/DocumentationButton/index.tsx
```diff
@@ -0,0 +1,36 @@
+
```
faouziH21, line 1:
```suggestion

```

src/components/DocumentationButton/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  className
+}) => {
+
+  const finalClassName = className
+    ? `${defaultClassName} ${className}`
+    : defaultClassName;
+
+  return (
+    <SimpleButton
+      onClick={() => window.open('/gis-docs', '_blank')}
```
marcjansen, line 23: The URL should be configurable, I guess.

src/components/Header/index.tsx
```diff
@@ -97,6 +100,25 @@ export const Header: React.FC<HeaderProps> = ({
...
+        key="documentation-button"
+        aria-label="documentation-button"
+      >
+        <DocumentationButton
+          key="documentation-button"
+          type="link"
+        >
+        </DocumentationButton>
+
+      </div>
```
marcjansen, line 115: Why do we need a wrapping `<div>`?

src/components/Header/index.tsx
```diff
@@ -97,6 +100,25 @@ export const Header: React.FC<HeaderProps> = ({
...
+    ];
+    insertPlugins('right', items);
+    if (ClientConfiguration.documentationButtonVisible) {
```
marcjansen, line 118: I'd move the check to the top of the fiunction:

```
if (!ClientConfiguration.documentationButtonVisible) {
  return;
}
```

marcjansen (commented):
Just some comments, others should also weigh in

src/components/Header/index.tsx
faouziH21: other components used a wrapping `<div>` hence I chose to also wrap it in a div.

src/components/Header/index.tsx
faouziH21: check [8861c02](https://github.com/terrestris/shogun-gis-client/pull/1743/commits/8861c02552193e756d49bccfefb9548435aadd86)

dnlkoch (approved):
Thanks @faouziH21!

PR #1748, opened 2024-09-16
feat: masks a link if present

FritzHoing:


buehner:
:+1: (once all tests are green)

buehner (approved):

src/components/DisplayField/index.tsx
```diff
@@ -135,7 +135,7 @@ export const DisplayField: React.FC<DisplayFieldProps> = ({
           target="_blank"
           rel='noreferrer'
         >
-          {value?.toString()}
+          Link
```
dnlkoch, line 138: This should be translatable.

PR #1749, opened 2024-09-17
Masked-URL

FritzHoing:
This MR masks an URL if a property `maskedUrl` is set within the `featureInfoFormConfig`. 
If that Property is not set the default value will be shown.

### DIFF
--- a/src/components/DisplayField/index.spec.tsx
+++ b/src/components/DisplayField/index.spec.tsx
@@ -107,7 +107,7 @@ describe('<DisplayField />', () => {
 
     for (const linkValue of linkValues) {
       render(<DisplayField value={linkValue} />);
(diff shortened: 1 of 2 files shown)


src/i18n/translations.ts
```diff
@@ -1,3 +1,5 @@
+import DisplayField from "components/DisplayField";
```
KaiVolland, line 1: I think this is an accident.

KaiVolland (commented):
Nice fix. But i think the translations.ts does not look like you want it to. Also the commit message could describe the fix better.

KaiVolland (approved):

dnlkoch (approved):
Thanks @FritzHoing!

PR #1758, opened 2024-11-05
`AddLayerModal`: Show layer in tree after adding it to the map

dnlkoch:
If the first layer is added to the map in the `AddLayerModal` (and the "External layers" folder is not present in the tree yet) the layer isn't visible in this folder. This fixes this issue by adding the layer to the target group before adding it to the map (to make all event listeners in the react-geo layer tree work as needed).


LukasLohoff (approved):

dnlkoch:
Thanks for the review!

PR #1759, opened 2024-11-06
Init `ImportDataModal` and fix minor styling issues

dnlkoch:
This replaces the `UploadButton` with the newly introduced `ImportDataModal` to actually allow to import GeoJSON files in the draw tools via drag & drop.

In addition some minor styling issues in the measure and draw toolbar emerged through the dependency updates were fixed.


marcjansen:
> null

:thinking: I thought we had this fixed...

dnlkoch:
> 🤔 I thought we had fixed this thingy...

I assume it's only fixed on `main`, but not on `next`.

LukasLohoff (approved):

dnlkoch:
Thanks for the review!

PR #1760, opened 2024-11-07
Update global loading mask

dnlkoch:
This updates the loading mask by:

- removing the static `Please wait, loading…` since it's not translatable
- adding an animated loading indicator instead
- cleaning up the css

Peek 2024-11-07 10-56


LukasLohoff (approved):
Looks really good now!

Unrelated to these changes: I think the fade-away animation would look better if the [shogun_spinner.gif](https://github.com/terrestris/shogun-gis-client/blob/main/resources/public/shogun_spinner.gif) had a transparent background.

KaiVolland:
IMHO the loading indicator is not needed as the leaf already has a spinning cog.

dnlkoch:
> Unrelated to these changes: I think the fade-away animation would look better if the [shogun_spinner.gif](https://github.com/terrestris/shogun-gis-client/blob/main/resources/public/shogun_spinner.gif) had a transparent background.

Good point! I tried to change the background, but couldn't fix it :-/ 

> IMHO the loading indicator is not needed as the leaf already has a spinning cog.

Yes, the default one is already animated, but the logo is usually changed in instances (to non-animated ones).

dnlkoch:
Thanks for the reviews!

PR #1762, opened 2024-11-08
fix: geometry edit tools and styling

AmandaTamanda:
Sets geometry edit tools as inactive until activated by the users and applies styling fixes.

### DIFF
--- a/src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
+++ b/src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
@@ -34,7 +34,7 @@ import { DeleteButton } from '@terrestris/react-geo/dist/Button/DeleteButton/Del
 import DrawButton from '@terrestris/react-geo/dist/Button/DrawButton/DrawButton';
 import { ModifyButton } from '@terrestris/react-geo/dist/Button/ModifyButton/ModifyButton';
 import SimpleButton from '@terrestris/react-geo/dist/Button/SimpleButton/SimpleButton';
-import ToggleGroup from '@terrestris/react-geo/dist/Button/ToggleGroup/ToggleGroup';
+import { ToggleGroup } from '@terrestris/react-geo/dist/Button/ToggleGroup/ToggleGroup';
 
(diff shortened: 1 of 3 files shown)


src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
```diff
@@ -221,10 +222,16 @@ export const EditFeatureGeometryToolbar: React.FC<EditFeatureGeometryToolbarProp
         className="geometry-edit-tb"
         alignment="vertical"
       >
-        <ToggleGroup>
+        <ToggleGroup
+        selected={selected}
+        onChange={(_evt: any, value: string | undefined) => {
+          setSelected(value)
+        }}
+        >
```
KaiVolland, line 230:
```suggestion
        <ToggleGroup
          selected={selected}
          onChange={(_evt: any, value: string | undefined) => {
            setSelected(value)
          }}
        >
```

src/components/EditFeatureDrawer/EditFeatureGeometryToolbar/index.tsx
```diff
@@ -221,10 +222,16 @@ export const EditFeatureGeometryToolbar: React.FC<EditFeatureGeometryToolbarProp
...
           {
             allowedEditMode.includes('CREATE') ?
               <DrawButton
+                value="one"
```
KaiVolland, line 234: IMHO a more speaking value like "CREATE" would be better. Same for the buttons below.

KaiVolland (commented):
Very nice. It's always nice to have a screenshot when styling changes if possible.

AmandaTamanda:
> Very nice. It's always nice to have a screenshot when styling changes if possible.

thanks for the review, the styling changes were more about restoring old styles which were no longer available due to the updates in the next version. I still generally looks the same.

KaiVolland (approved):

dnlkoch (approved):
Very nice, thanks!

AmandaTamanda:
> Very nice, thanks!

Thanks for the review!

PR #1764, opened 2024-11-15
Introduce WFS search

dnlkoch:
This includes a generic WFS search option to the `MultiSearch` component while refactoring it to become no longer strongly linked to specific search engines.

Peek 2024-11-28 14-44

To activate the search engine it's sufficient to add it to the `engines` array in the search config of the selected instance:

```json
{
  "name": "search",
  "config": {
    "engines": [
      "nominatim",
      "wfs"
    ]
  }
}
```

For a searchable layer the existing configuration is untouched and supports `solr` and `wfs` the same way:

```json
{
  "searchable": true,
  "searchConfig": {
    "attributes": [
      "NAME",
      "scalerank"
    ],
    "displayTemplate": "Result: {NAME}"
  }
}
```

Additional notes: 

- this also suggests to removes the `useNominatim` option from the global application configurations since it's configurable by instance anyway.
- it's now possible to configure the nominatim server URL globally.
- the existing `BasicNominatimSearch` has been removed in favour of the refactored `MultiSearch` (having the same behaviour, but enables users to limit the search results to the given extent). 


hwbllmnn (approved):

dnlkoch:
Thanks for the review!

PR #1771, opened 2024-11-22
feat: adds wms styling using geostyler

FritzHoing:
This MR adds wms styling via the geostyler using the SLD_BODY.

wms-geostyler

Please note:
The geostyler has a css issue in the current version, that needs to be fixed. The canvas of the style is dislocated.

### DIFF
--- a/src/components/StylingDrawer/SldStylingPanel/index.tsx
+++ b/src/components/StylingDrawer/SldStylingPanel/index.tsx
@@ -1,0 +1,109 @@
+import React, {
+  useEffect, useState, useCallback
+} from 'react';
+
+import {
+  CardStyle, CardStyleProps
+} from 'geostyler/dist/Component/CardStyle/CardStyle';
+import SLDParser from 'geostyler-sld-parser';
+import { Style as GsStyle } from 'geostyler-style';
+import Layer from 'ol/layer/Layer';
+import {
+  ImageWMS, TileWMS
+} from 'ol/source';
+
+import { Logger } from '@terrestris/base-util';
+import { MapUtil } from '@terrestris/ol-util';
+import { useMap } from '@terrestris/react-util/dist/Hooks/useMap/useMap';
+
+import useAppSelector from '../../../hooks/useAppSelector';
+import useSHOGunAPIClient from '../../../hooks/useSHOGunAPIClient';
+import {
+  fetchGeoserverStyle, fetchWorkspaceFromGetCapabilities, getLayerUrl
+} from '../../../utils/geoserverUtils';
+
+export type SldStylingPanelProps = CardStyleProps;
+
+const SldStylingPanel: React.FC<SldStylingPanelProps> = (props): JSX.Element => {
+  const [style, setStyle] = useState<GsStyle | undefined>();
+  const layerUid = useAppSelector(state => state.stylingDrawerLayerUid);
+  const map = useMap();
+  const client = useSHOGunAPIClient();
+
+  const getStyle = useCallback(async () => {
+    if (!map || !layerUid) {
+      return;
+    }
+
+    const layer = MapUtil.getLayerByOlUid(map, layerUid) as Layer;
+    if (!layer) {
+      return;
+    }
+
+    const source = layer.getSource();
+    if (!(source instanceof ImageWMS || source instanceof TileWMS)) {
+      return;
+    }
+
+    const layerUrl = getLayerUrl(source);
+    if (!layerUrl) {
+      return;
+    }
+
+    try {
+      const workspaceInfo = await fetchWorkspaceFromGetCapabilities(layerUrl, client);
+      if (!workspaceInfo) {return;}
+
+      const geoserverStyle = await fetchGeoserverStyle(workspaceInfo.workspace, workspaceInfo.layerName, layerUrl, client);
+      if (geoserverStyle) {
+        const parser = new SLDParser();
+        const { output: sldObject } = await parser.readStyle(geoserverStyle);
+        setStyle(sldObject);
+      }
+    } catch (error) {
+      Logger.error('Error: ', error);
+    }
+  }, [map, layerUid, client]);
+
+  useEffect(() => {
+    getStyle();
+  }, [getStyle]);
+
+  useEffect(() => {
+    if (!map || !style) {
+      return;
+    }
+
+    const layer = MapUtil.getLayerByOlUid(map, layerUid) as Layer;
+    const source = layer?.getSource() as ImageWMS;
+    if (!source) {
+      return;
+    }
+
+    style.name = source.getParams().LAYERS;
+
+    const parser = new SLDParser();
+    parser.writeStyle(style).then(({ output: sld }) => {
+      if (sld) {
+        source.updateParams({
+          SLD_BODY: sld,
+          STYLES: style.name
+        });
+      }
+    });
+  }, [style, map, layerUid]);
+
+  if (!map || !style) {
+    return <></>;
+  }
+
+  return (
+    <CardStyle
+      style={style}
+      onStyleChange={setStyle}
+      {...props}
+    />
+  );
+};
+
+export default SldStylingPanel;
--- a/src/components/StylingDrawer/index.tsx
+++ b/src/components/StylingDrawer/index.tsx
@@ -1,20 +1,33 @@
-import React from 'react';
+import React, {
+  useEffect, useState
+} from 'react';
 
 import {
   Drawer,
   DrawerProps
 } from 'antd';
 
+import Layer from 'ol/layer/Layer';
+import {
+  ImageWMS, TileWMS
+} from 'ol/source';
 import {
   useTranslation
 } from 'react-i18next';
 
 import './index.less';
 
+import { MapUtil } from '@terrestris/ol-util';
+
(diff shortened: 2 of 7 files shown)


src/utils/geoserverUtils.tsx
```diff
@@ -0,0 +1,123 @@
...
+  if (layerSource instanceof TileWMS || layerSource instanceof ImageWMS) {
+    url = layerSource instanceof TileWMS ? layerSource.getUrls()?.[0] : layerSource.getUrl();
+  }
+  return url || false;
+};
+
+export const checkIfGeoserverLayer = (layerSource: Source | undefined): boolean => {
+  const url = getLayerUrl(layerSource);
+  return url ? url.includes('/geoserver/') : false;
+};
```
dnlkoch, line 121: This check is quite unreliable since there a GeoServers out there that aren't served from the default path.

src/utils/geoserverUtils.tsx
```diff
@@ -0,0 +1,123 @@
...
+    return undefined;
+  }
+};
+
+export const getLayerUrl = (layerSource: Source | undefined): string | false => {
+  if (!layerSource) {
+    return false;
+  }
+  let url: string | undefined;
+  if (layerSource instanceof TileWMS || layerSource instanceof ImageWMS) {
```
dnlkoch, line 112: Very minor, but we might want to reuse the [isWmsLayer](https://github.com/terrestris/ol-util/blob/main/src/typeUtils/typeUtils.ts#L17) typeguard instead.

src/utils/geoserverUtils.tsx
```diff
@@ -0,0 +1,123 @@
...
+    url = layerSource instanceof TileWMS ? layerSource.getUrls()?.[0] : layerSource.getUrl();
+  }
+  return url || false;
```
dnlkoch, line 115: I would assume that it's very uncommon to return `false` here. If no url can be found, we should just return `undefined` or `null`.

src/utils/geoserverUtils.tsx
```diff
@@ -0,0 +1,123 @@
...
+  } catch (error) {
+    Logger.error('Error: ', error);
+    return undefined;
+  }
+};
+
+export const getLayerUrl = (layerSource: Source | undefined): string | false => {
+  if (!layerSource) {
+    return false;
+  }
```
dnlkoch, line 110: This check can be omitted if we would make the input argument mandatory.

src/utils/geoserverUtils.tsx
```diff
@@ -0,0 +1,123 @@
...
+};
+
+export const fetchGeoserverStyle = async (
+  layerWorkspace: string,
+  layerName: string,
+  layerUrl: string,
+  client: SHOGunAPIClient | null
+): Promise<string | undefined> => {
+  const cleanLayerUrl = layerUrl.replace('/ows?', '');
+  const url = `${cleanLayerUrl}/rest/layers/${layerWorkspace}:${layerName}.json`;
```
dnlkoch, line 84: This URL (and also the one used in `fetchSld()`) is usually accessible for users holding the role `ROLE_ADMINISTRATOR` in GeoServer only.

src/utils/geoserverUtils.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  try {
+    const response = await fetch(url, {
+      headers: {
+        ...getBearerTokenHeader(client?.getKeycloak())
+      }
```
dnlkoch, line 90: This should be passed to internal layers (identified by `useBearerToken`) only.

src/components/StylingDrawer/SldStylingPanel/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+      return;
+    }
+
+    const layerUrl = getLayerUrl(source);
+    if (!layerUrl) {
+      return;
+    }
+
+    try {
+      const workspaceInfo = await fetchWorkspaceFromGetCapabilities(layerUrl, client);
```
dnlkoch, line 54: Just wondering why this is needed. Usually the workspace is already party of the layer name, e.g. `MYWORKSPACE:MYLAYER`. This can probably enhanced by checking if the name contains the workspace already and if not getting it from the capabilities.

src/components/StylingDrawer/SldStylingPanel/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    style.name = source.getParams().LAYERS;
+
+    const parser = new SLDParser();
+    parser.writeStyle(style).then(({ output: sld }) => {
+      if (sld) {
+        source.updateParams({
+          SLD_BODY: sld,
+          STYLES: style.name
+        });
```
dnlkoch, line 91: Just wondering, but what happens if:

- a layer has been opened in the drawer,
- got a new style via `SLD_BODY`,
- the drawer has been closed and reopened with the same layer.

What style would be shown in the styler UI? The default one or the updated one?

src/components/StylingDrawer/index.tsx
```diff
@@ -1,36 +1,74 @@
...
   const isStylingDrawerVisible = useAppSelector(state => state.stylingDrawerVisibility);
+  const layerUid = useAppSelector(state => state.stylingDrawerLayerUid);
+  const map = useMap();
 
   const {
     t
   } = useTranslation();
 
+  useEffect(() => {
+
```
dnlkoch, line 49:
```suggestion
```

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -293,6 +316,18 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
     });
   }
 
+  if (isGeoserverLayer) {
+    dropdownMenuItems.push({
+      label: 'Geostyler',
```
dnlkoch, line 321: The label should be translatable and the current one might be confusing to users. We might want to replace it with something like `Layer style`.

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -293,6 +316,18 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
...
+      key: 'geostyler'
+    });
+  }
+
+  dropdownMenuItems.push({
+    label: t('LayerTreeContextMenu.layerDetails'),
+    key: 'layerDetails'
+  });
```
dnlkoch, line 329: Why is this needed? This will be added further down (conditionally).

src/components/ToolMenu/LayerTree/LayerTreeContextMenu/index.tsx
```diff
@@ -238,6 +256,11 @@ export const LayerTreeContextMenu: React.FC<LayerTreeContextMenuProps> = ({
     a.click();
   };
 
+  const configureStyles = () => {
+    dispatch(setStylingDrawerVisibility(true));
+    dispatch(setStylingDrawerLayerUid(getUid(layer)));
```
dnlkoch, line 261: Should we show the styler if there is no layer present?

dnlkoch (changes requested):
Thank you very much for this addition @FritzHoing! 

Overall the functionality is already awesome and is working for a defined context obviously. But for the production scenarios I think we need to adjust some architectural assumptions before we can merge this:

1. Guessing the layer's map server by its path is very prone to errors since it might not match non-default scenarios. Probably it's the easiest way to just make the style-function configurable by the layer, e.g. by introducing a property in the layer configuration like `styleable` (to match the existing keys like `searchable` or `editable`). A `styleable` layer might then be defined as layer coming from GeoServer (this is open for discussion of course) or we might add an additional config like `styleConfig` containing the URL where to get the default style from.
2. Using the _internal_ GeoServer REST interfaces to get the style is generally fine, but in default scenarios this is accessible for admin users only. Default app users won't be able to access them. An easy approach would be to introduce a new _proxy_ interface for the style URLS in the SHOGun backend providing access to them. An alternate approach could be using the `WMS GetStyles` request, but I'm not sure if this is actually supported as expected by GeoServer.

But again, I really like how this is looking so far and it would be really awesome to have this included!

src/components/StylingDrawer/SldStylingPanel/index.tsx
FritzHoing: Good hint! The default style will be used in that case. Maybe we want to check if a layer has a SLD_BODY already?

src/utils/geoserverUtils.tsx
FritzHoing: Yes i am aware of that issue, but i don't know how else i could identify if a layer is served by a geoserver.

src/components/StylingDrawer/SldStylingPanel/index.tsx
FritzHoing: Maybe we also want to store the `SLD_BODY`?

src/utils/geoserverUtils.tsx
dnlkoch: Yes, indeed, this is not easy to guess and I think the only way is to actually set a identifier to the layer (see my suggestions in the review comment).

PR #1772, opened 2024-11-29
Shogun gis client layout fixes

AmandaTamanda:
Bildschirmfoto vom 2024-11-29 10-08-17

Bildschirmfoto vom 2024-11-29 10-08-48

### DIFF
--- a/src/components/Header/index.less
+++ b/src/components/Header/index.less
@@ -20,7 +20,7 @@
       justify-content: flex-start;
 
       .logo {
-        height: 100%;
+        height: 80%;
         padding: 10px;
       }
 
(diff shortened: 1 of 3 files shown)


src/components/Header/index.less
```diff
@@ -20,7 +20,7 @@
       justify-content: flex-start;
 
       .logo {
-        height: 100%;
+        height: 80%;
```
dnlkoch, line 23: There is a padding set already. Is this not working as expected?

dnlkoch (approved):

src/components/Header/index.less
AmandaTamanda: No, it is not: The padding is only applied left and right when increased

AmandaTamanda: tried with the margin, same problem. I can however adjust the padding on the parent div instead. Would that be a more suitable solution?

PR #1773, opened 2024-11-29
fix: adds padding and extends link to the text

AmandaTamanda:
image


dnlkoch (commented):
Looks good, but I would favour to have the same padding at all sides of the boundary.

PR #1774, opened 2024-12-03
Readd support for a custom favicon path

dnlkoch:
Since the we don't set the deprecated `shortcut icon` anymore in the index template, the selector needs to be updated as well.


buehner (approved):

dnlkoch:
Thanks for the review!

PR #1775, opened 2024-12-03
Fix layer loading indicator in tree

dnlkoch:


dnlkoch:
See [#895](https://github.com/terrestris/shogun-util/pull/895) as well.

PR #1782, opened 2024-12-18
Fix layer loading indicator in tree

dnlkoch:
Create deep clone from state value before manipulating it.

Extracted from #1775.


ahennr (approved):

dnlkoch:
Thanks for the review!

PR #1783, opened 2024-12-18
fix: layertree checkbox css

FritzHoing:
This MR fixes the CSS of the layertree checkboxes.
The initial CSS got changed in antd so it needs to be overwritten using `!important`

image

### DIFF
--- a/src/components/ToolMenu/LayerTree/index.less
+++ b/src/components/ToolMenu/LayerTree/index.less
@@ -79,3 +79,8 @@
     }
   }
 }
+
+.tree-wrapper .ant-tree-checkbox {
+  top: 2.5px;
+  align-self: flex-start !important;
+}


src/components/ToolMenu/LayerTree/index.less
```diff
@@ -79,3 +79,8 @@
     }
   }
 }
+
+.ant-tree-checkbox {
```
dnlkoch, line 83: This will override all `ant-tree-checkbox` classes in the application. Is this intended?

FritzHoing: You are right! Due to the antd change we might want that actually if the checkboxes seem to be dislocated in other components too. But for this case i made the selector more specific.

dnlkoch (approved):

PR #1784, opened 2024-12-18
feat: adds Map Toolbar to the gis-client

AmandaTamanda:


src/i18n/translations.ts
```diff
@@ -126,6 +126,11 @@ export default {
         layertree: 'Karten',
         languageSelect: 'Sprachauswahl'
       },
+      MapToolbar: {
+        zoomInTooltip: 'Reinzoomen',
+        zoomOutTooltip: 'Rauszoomen',
```
LukasLohoff, line 131:
```suggestion
        zoomInTooltip: 'Hereinzoomen',
        zoomOutTooltip: 'Herauszoomen',
```

LukasLohoff (approved):

AmandaTamanda:
Thanks for the review @LukasLohoff

src/components/MapToolbar/index.less
```diff
@@ -0,0 +1,21 @@
...
+
+  .ant-btn {
+    margin: 5px;
+    display: block;
+    color: var(--complementaryColor);
+    background: var(--primaryColor);
+    border-color: var(--primaryColor);
+
+    &:hover {
+      background: var(--secondaryColor) !important;
```
dnlkoch, line 17: We should not use `!important`. Can we define a more specific selector to get rid of it?

src/components/MapToolbar/index.less
```diff
@@ -0,0 +1,21 @@
+.map-toolbar {
+  position: absolute;
+  top: calc(var(--headerHeight) + 1em);
+  right: 10px;
```
dnlkoch, line 4: Is there any reason why we mix `em` and `px` units here?

src/components/MapToolbar/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export type MapToolbarProps = React.HTMLAttributes<HTMLDivElement> & ToolbarProps & {};
+
+export const MapToolbar: React.FC = (): JSX.Element => {
+  const {
+    t
+  } = useTranslation();
+  const map = useMap();
+
+  const [pressed, setPressed] = useState(false);
```
dnlkoch, line 43: Very minor, but since this is only for the geolocation button it might be more intuitive to rename the variable to something like `geolocationButtonIsPressed`.

src/components/MapToolbar/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const [visible] = useState(store.getState().toolMenu.availableTools.find((toolName) => toolName === 'map_toolbar'));
```
dnlkoch, line 45: There is no need to create a state variable here and the state of the store should always be accessed via the [useAppSelector](https://github.com/terrestris/shogun-gis-client/blob/next/src/hooks/useAppSelector.ts) hook like [here](https://github.com/terrestris/shogun-gis-client/blob/next/src/components/ToolMenu/index.tsx#L114).

src/bootstrap.tsx
```diff
@@ -290,6 +293,9 @@ const setApplicationToStore = async (application?: Application) => {
         if (tool.name === 'tree' && typeof tool.config.layerIconsVisible !== 'undefined') {
           store.dispatch(setLayerIconsVisible(tool.config.layerIconsVisible));
         }
+        if (tool.name === 'map_toolbar') {
+          store.dispatch(setMapToolbarVisible(tool.config.visible));
```
dnlkoch, line 297: What happens if `tool.config` is `null` or `undefined`?

dnlkoch: And just wondering if this is needed anyway? In the `MapToolbar` we're just reading out the value from `availableTools`.

src/components/MapToolbar/index.less
AmandaTamanda: Fixed

src/components/MapToolbar/index.tsx
AmandaTamanda: done

src/components/MapToolbar/index.less
AmandaTamanda: I have changed it to uniform `px`

src/bootstrap.tsx
AmandaTamanda: @dnlkoch I am not sure  I understand, the dispatch is not needed here?

src/components/MapToolbar/index.tsx
AmandaTamanda: Thank you, I have made the necessary changes.

PR #1785, opened 2024-12-19
feat: adds tooltips to layertree icons

AmandaTamanda:
image


src/i18n/translations.ts
```diff
@@ -420,7 +423,10 @@ export default {
         uploadData: 'Upload data',
         print: 'Export',
         layertree: 'Maps',
-        languageSelect: 'Language selector'
+        languageSelect: 'Language selector',
+        searchable: 'searchable',
+        queryable: 'queryable',
+        editable: 'editable'
```
dnlkoch, line 429: The hints could be a bit more intuitive, e.g. something like:
- `searchable`: _The contents of this layer can be queried in the search input_
- `queryable`: _The contents of this layer can be queried in the query map features tool_
- `editable`: _The features of this layer are editable_

FritzHoing: Thank you for the review! The translations are adjusted according to your suggestions.

dnlkoch (approved):

PR #1787, opened 2024-12-23
Show legend for WMS layers only

dnlkoch:
This fixes an unchecked cast which might result in an error since the `Legend` component does accept WMS layers only.


ahennr (approved):

dnlkoch:
Thanks for the review!

PR #1788, opened 2024-12-23
Reactivate drill down feature info requests

dnlkoch:
This (re)enables `drillDown` in the `FeatureInfo` component thus querying all stacked layers.


ahennr (approved):

dnlkoch:
Thanks for the review!

PR #1789, opened 2024-12-23
Prevent unnecessary rerenders of the legend

dnlkoch:
This prevents the `Legend` from unneeded rerenders (and annoying loading indicators in it) by caching the props `extraParams` and `headers`.

Before:

Peek 2024-12-23 13-33

After:

Peek 2024-12-23 13-31


ahennr (approved):

dnlkoch:
Thanks for the review!

PR #1823, opened 2025-01-30
Move application info opener to footer

dnlkoch:
This suggests to move the application info opener ("About") from the user menu to the footer. This allows users to open it regardless from the presence of the user menu which visible state is defined as follows now:

- visible by default always, except (possibly required for public instances):
  - not visible if Keycloak is disabled globally (see `keycloak.enabled`) -> no change in behaviour
  - not visible if explicitly disabled for the given application configuration -> new configuration:
  ```
  {
    "name": "user_menu",
    "config": {
      "visible": false
    }
  }
  ```

The menu itself is slightly refactored to show a login button only if a user is not logged in already and to show the users full name as well. Previously the username was shown only.

login

image

In addition the `DocumentationButton` is slightly simplified.

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -129,6 +129,9 @@ import {
 import {
   setUser
 } from './store/user';
+import {
+  setVisible as setUserMenuVisible
+} from './store/userMenu';
 
 import './index.less';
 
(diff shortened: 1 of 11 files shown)


ahennr (approved):

dnlkoch:
Thanks for the review!

PR #1833, opened 2025-02-04
Fix several code smells detected by SonarQube

marcjansen:
Title says it all, please review.

### DIFF
--- a/src/components/ApplicationInfo/index.tsx
+++ b/src/components/ApplicationInfo/index.tsx
@@ -39,10 +39,10 @@ export const ApplicationInfo: React.FC<ApplicationInfoProps> = ({
   const logoPath = useAppSelector(state => state.logoPath);
   const appDescription = useAppSelector(state => state.description);
 
-  const [isVisible, setVisible] = useState<boolean>(false);
+  const [isVisible, setIsVisible] = useState<boolean>(false);
 
   const toggleVisibility = () => {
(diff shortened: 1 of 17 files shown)


src/components/ApplicationInfo/index.tsx
```diff
@@ -39,10 +39,10 @@ export const ApplicationInfo: React.FC<ApplicationInfoProps> = ({
   const logoPath = useAppSelector(state => state.logoPath);
   const appDescription = useAppSelector(state => state.description);
 
-  const [isVisible, setVisible] = useState<boolean>(false);
+  const [visible, setVisible] = useState<boolean>(false);
```
dnlkoch, line 42: Personally I would prefer `isVisible`. Probably this can be fixed by renaming the setter to `setIsVisible`?

marcjansen: I think so, and can do this

marcjansen: done in dd02a0b4e3316131a38e3fffea6b5dc527b3077a

dnlkoch:
And can we squash the commits before merging?

marcjansen:
> And can we squash the commits before merging?

I can do that as well. All in one commit, or grouped by type of change?

dnlkoch:
A single commit is just fine.

marcjansen:
Thanks for the review

PR #1836, opened 2025-02-06
Legend: Set current map scale on initial render

dnlkoch:
Enabling the legend (in the layer tree) without having moved the map before might result in an erroneous or inaccurate `GetLegendGraphic` request since the map scale is `undefined`. 

This suggests to fix this issue by setting the default scale value to the current map scale. 


LukasLohoff (approved):

dnlkoch:
Thanks for the review!

PR #1857, opened 2025-02-13
test: mock react-i18next globally

sdobbert:
This mocks react-i18next in the jest-setup.


dnlkoch:
Could please change the target branch to the `main`?

dnlkoch (approved):

PR #1880, opened 2025-02-20
test: unittest for StylingComponent

sdobbert:
Adds a unittest for the StylingComponent.

### DIFF
--- a/src/components/ToolMenu/Draw/StylingDrawerButton/StylingComponent/index.spec.tsx
+++ b/src/components/ToolMenu/Draw/StylingDrawerButton/StylingComponent/index.spec.tsx
@@ -44,10 +44,6 @@ describe('StylingDrawer', () => {
     (MapUtil.getLayerByName as jest.Mock).mockReturnValue(mockVectorLayer);
   });
 
-  it('is defined', () => {
-    expect(StylingComponent).not.toBeUndefined();
-  });
-
   it('can be rendered', () => {
     const {
       container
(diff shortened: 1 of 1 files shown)


src/components/ToolMenu/Draw/StylingDrawerButton/StylingComponent/index.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+      getLayers: jest.fn()
+    });
+
+    const mockVectorLayer = new OlLayerVector({ source: new OlSourceVector() });
+    (MapUtil.getLayerByName as jest.Mock).mockReturnValue(mockVectorLayer);
+  });
+
+  it('is defined', () => {
+    expect(StylingComponent).not.toBeUndefined();
+  });
```
dnlkoch, line 49: Non-blocking, but generally I think these kind of tests aren't needed.

src/components/ToolMenu/Draw/StylingDrawerButton/StylingComponent/index.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+    expect(inputElement).toBeInTheDocument();
+    expect(inputElement.value).toBe('Default Style');
+  });
+
+  it('processes styles and applies them to the map', async () => {
+    render(<StylingComponent />);
+
+    await waitFor(() => {
+      expect(OlParser).toHaveBeenCalled();
+      expect(MapUtil.getLayerByName).toHaveBeenCalledWith(expect.any(Object), 'react-geo_digitize');
```
dnlkoch, line 72: Can we also check if the layer has been added to the map actually?

dnlkoch (approved):

src/components/ToolMenu/Draw/StylingDrawerButton/StylingComponent/index.spec.tsx
sdobbert: Thanks, I removed this test.

PR #1943, opened 2025-03-18
Introduces SearchResultDrawer

KaiVolland:
This introduces the SearchResultDrawer.

The feature can be enabled when setting [searchResultDrawer](https://github.com/terrestris/shogun-gis-client/compare/search-result-drawer?expand=1#diff-858566d2d4cf06579a908cb85f587c5752fa0fa6a47d579277749006e86f0834R38) to true in the `SearchConfig` of the gis-client-config.js.

When set to true the search result wil be displayed in a drawer on the right side.

It contains the "title" property of the selected feature of the `SearchResultList`.

When the passed feature has other properties then title a `PropertyGrid` with its features is shown.

TODO:

- [x] Adapt `useSolrSearchEngine` to add all properties of the selected item to the created ol feature
- [x] Adapt `useWfsSearchEngine` to add all properties of the selected item to the created ol feature
- [x] Make PropertyGrid configurable
  - [x] attributesFilter (maybe reusable from layer config)
  - [x] Check if we want a more versatile configuration here

## Preview

shogun-client-search-result-drawer

### DIFF
--- a/src/components/SearchResultDrawer/index.tsx
+++ b/src/components/SearchResultDrawer/index.tsx
@@ -1,0 +1,157 @@
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState
+} from 'react';
+
+import {
+  DrawerProps
+} from 'antd';
+
+import OlFeature from 'ol/Feature';
+import GeoJSONParser from 'ol/format/GeoJSON';
+import OlLayerVector from 'ol/layer/Vector';
+import OlSourceVector from 'ol/source/Vector';
+import OlStyleCircle from 'ol/style/Circle';
+import OlStyleFill from 'ol/style/Fill';
+import OlStyleStroke from 'ol/style/Stroke';
+import OlStyle from 'ol/style/Style';
+
+import {
+  useTranslation
+} from 'react-i18next';
+
+import PropertyGrid from '@terrestris/react-geo/dist/Grid/PropertyGrid/PropertyGrid';
+import { useMap } from '@terrestris/react-util/dist/Hooks/useMap/useMap';
+
+import useAppDispatch from '../../hooks/useAppDispatch';
+import useAppSelector from '../../hooks/useAppSelector';
+import {
+  setSearchResultState
+} from '../../store/searchResult';
+import MapDrawer from '../MapDrawer';
+
+export type SearchResultDrawerProps = DrawerProps;
+
+export const SearchResultDrawer: React.FC<SearchResultDrawerProps> = ({
+  ...passThroughProps
+}): JSX.Element => {
+
+  const dispatch = useAppDispatch();
+  const map = useMap();
+  const isSearchResultDrawerVisible = useAppSelector(state => state.searchResult.drawerVisibility);
+  const geoJSONFeature = useAppSelector(state => state.searchResult.geoJSONFeature);
+  const [highlightLayer, setHighlightLayer] = useState<OlLayerVector<OlSourceVector> | null>(null);
+  const olFeature: OlFeature | null = useMemo(() => {
+    if (!geoJSONFeature) {
+      return null;
+    }
+    return new GeoJSONParser().readFeature(geoJSONFeature) as OlFeature;
+  }, [geoJSONFeature]);
+
+  useEffect(() => {
+    if (!map) {
+      return;
+    }
+
+    const layer = new OlLayerVector({
+      source: new OlSourceVector(),
+      style: new OlStyle({
+        stroke: new OlStyleStroke({
+          color: 'rgb(255,0,0)',
+          width: 2
+        }),
+        fill: new OlStyleFill({
+          color: 'rgba(255,255,255, 0.5)'
+        }),
+        image: new OlStyleCircle({
+          radius: 10,
+          fill: new OlStyleFill({
+            color: 'rgba(255,255,255, 0.5)'
+          }),
+          stroke: new OlStyleStroke({
+            color: 'rgb(255,0,0)',
+            width: 3
+          })
+        })
+      })
+    });
+
+    setHighlightLayer(layer);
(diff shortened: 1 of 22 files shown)


mholthausen:
The component has now been improved so that, in addition to the nominatim search, it also displays the features in a search via WFS or SOLR in the drawer. 

The grid in the drawer has been made configurable for WFS and SOLR searches so that it can be configured accordingly via the shogun admin (`searchFeatureConfig`) and displays custom defined attribute names instead of the keys. 

URLs can be displayed in abbreviated form using their own defined link text. The title can be a template that is displayed as title.

src/components/SearchResultDrawer/index.tsx
```diff
@@ -141,7 +195,35 @@ export const SearchResultDrawer: React.FC<SearchResultDrawerProps> = ({
...
+                      href={normalized}
+                      target="_blank"
+                    >
+                      {t('FeaturePropertyGrid.linkText')}
+                    </a>
+                  );
+                }
+
+                return normalized ?? '';
+              }
```
KaiVolland: This might me be better in its own component to enhance readability.

src/global.d.ts
```diff
@@ -35,6 +35,7 @@ declare module 'clientConfig' {
     coreName?: string;
     solrQueryConfig?: SolrQueryConfig;
     activateLayerOnClick?: boolean;
+    searchResultDrawer?: boolean;
```
dnlkoch, line 38: I would recommend to rename the config to a more meaningful name expecting a boolean value like `useSearchResultDrawer`, `showSearchResultDrawer` or similiar.

src/hooks/searchEngines/useSolrSearchEngine.ts
```diff
@@ -105,6 +105,34 @@ export const useSolrSearchEngine = () => {
...
+
+    Object.keys(dsResult)
+      .filter(key => !blacklistedAttributes.includes(key))
+      .forEach(key => {
+        const value = dsResult[key];
+        if (value !== undefined) {
+          olFeature.set(key, value);
+        }
+      });
+  }, [map]);
```
dnlkoch, line 134: `map` doesn't seem to be required in the callback?

src/hooks/searchEngines/useWfsSearchEngine.ts
```diff
@@ -81,6 +81,26 @@ export const useWfsSearchEngine = () => {
...
 
+  const applyAttributesToFeature = (
+    feature: OlFeature
+  ): void => {
+    const blacklistedAttributes = [
+      'category',
+      'id',
+      'featureType',
+      'geometry',
+      'search'
```
dnlkoch, line 92: Where is this blacklist coming from? What happens if a feature actually contains the `category` attribute?

src/components/SearchResultDrawer/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+          stroke: new OlStyleStroke({
+            color: 'rgb(255,0,0)',
+            width: 3
+          })
+        })
+      })
+    });
+
+    setHighlightLayer(layer);
+    map.addLayer(layer);
```
dnlkoch, line 93: Do we need to remove the layer on unmount? We also have [this](https://github.com/terrestris/react-util/blob/main/src/Hooks/useOlLayer/useOlLayer.ts) hook for comparable scenarios.

src/components/SearchResultDrawer/index.tsx
mholthausen: Right, I have created `AttributeValueCell` which will be used here instead.

src/hooks/searchEngines/useSolrSearchEngine.ts
mholthausen: Yes, I have removed `map` here.

src/hooks/searchEngines/useWfsSearchEngine.ts
mholthausen: I have used the blacklist from the https://github.com/terrestris/shogun-gis-client/blame/next/src/hooks/searchEngines/useWfsSearchEngine.ts#L59 as a reference.  
Should we modify this list in any way?

src/components/SearchResultDrawer/index.tsx
mholthausen: Yes, the hook from react-util will be used here now.

src/global.d.ts
mholthausen: I have renamed this config to `showSearchResultDrawer` :+1:

src/hooks/searchEngines/useWfsSearchEngine.ts
dnlkoch: Ah, obviously I introduced it ;-)

Yes, I would suggest to simplify this code snippet to filter out just the geometry. Since this PR allows to define the properties visible in the UI, this shouldn't be a problem, right?

mholthausen: Right, as they can be configured now, I have removed all other fields but `geometry`.

src/store/searchResult/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  drawerVisibility: boolean;
+}
+
+const initialState: SearchResultState = {
+  drawerVisibility: false,
+  geoJSONFeature: null
+};
+
+export const slice = createSlice({
+  name: 'stylingDrawerVisibility',
```
dnlkoch, line 19:
```suggestion
  name: 'searchResultDrawer',
```

Not sure if this change requires other changes as well…

dnlkoch (approved):
Very nice @mholthausen and @KaiVolland!

Could you please check the commit history? It looks like some dependabot commits where included by mistake.

PR #2090, opened 2025-05-15
Fix pipeline for `next` branch without any new release

dnlkoch:
This attempts to fix the currently failing pipeline in the #1976 branch.

Explanation: As there is currently no release performed for the branch, the variable `release_version` is not defined and the checkout can't be performed. As fallback the previous version/tag is used now.


ahennr (approved):

simonseyock (approved):

dnlkoch:
Thanks for the reviews!

dnlkoch:
Unfortunately this is not working, `last_release_version` is not being set and it seems like an open issue in the [action](https://github.com/cycjimmy/semantic-release-action/issues/137).

PR #2099, opened 2025-05-19
Unify ol import paths for rspack

dnlkoch:
This pull request introduces a custom rspack plugin, `OlImportRewritePlugin`, to address issues with OpenLayers import paths and integrates it into the build configuration. The plugin ensures consistent import paths by removing the `.js` extension, preventing module duplication in module federation setups. Without unified import paths, the module federation is actually providing shared modules for the same module twice, e.g. sharing the module `ol/Map` and `ol/Map.js`. This is actually leading to potential scope (and e.g. resulting in `instanceof` check) issues during runtime.

### DIFF
--- a/OlImportRewritePlugin.js
+++ b/OlImportRewritePlugin.js
@@ -1,0 +1,25 @@
+/**
+ * This plugin rewrite all OpenLayers import paths by removing the '.js'
+ * extension from the import paths (e.g. 'ol/Map.js' to 'ol/Map').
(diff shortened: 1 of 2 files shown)


simonseyock (approved):
Thanks a lot! This might make the `CutDeleteButton` from react-geo actually work in plugin modules!

dnlkoch:
Thanks for the review!

PR #2115, opened 2025-05-26
feat: add possibility to deactivate map interaction tools programmatically

simonseyock:
If using a map interaction tool (like a geometry editor) from a plugin, the shogun-gis-clients map interaction tools stayed active and there was no easy way to deactivate them.

To handle this I introduced a `mapInteractionStatus` state variable that will deactivate the corresponding tool if set to a different value.

To test this I included a simple button in the footer of the application:

Peek 2025-05-26 16-49

The code for the button looks like this:

```tsx
import React, {useMemo} from "react";
import useAppDispatch from "./hooks/useAppDispatch";
import useAppSelector from "./hooks/useAppSelector";
import {MapInteractionStatus, removeInteractionStatus, setMapInteractionStatus} from "./store/mapInteractionStatus";
import {Button} from "antd";

export const TestMapInteraction = () => {
  const mapInteractionStatus = useAppSelector(state => state.mapInteractionStatus);
  const dispatch = useAppDispatch();
  const isActive = useMemo(() => mapInteractionStatus === ('other' as MapInteractionStatus), [mapInteractionStatus]);

  return <div>
    <Button
      type={'primary'}
      style={{
        backgroundColor: isActive ? 'green' : undefined
      }}
      onClick={() => {
        if (!isActive) {
          dispatch(setMapInteractionStatus('other' as MapInteractionStatus))
        }
        else {
          dispatch(removeInteractionStatus('other' as MapInteractionStatus))
        }
      }}>
      turn on other interaction
    </Button>
  </div>;
}
```

### DIFF
--- a/src/components/ToolMenu/Draw/index.tsx
+++ b/src/components/ToolMenu/Draw/index.tsx
@@ -1,5 +1,6 @@
 import React, {
   useCallback,
+  useEffect,
   useState
 } from 'react';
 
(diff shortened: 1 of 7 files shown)


simonseyock:
As disabling tools is possible via the `setActiveTools` action, this is not really necessary.

PR #2117, opened 2025-05-27
Fixes for the WFS search

dnlkoch:
This adds support for numeric attributes in the WFS search and removes the unneeded blacklisting of attributes (it seems to be a leftover of the solr search this one is based upon).


LukasLohoff (approved):
Looks good!

PR #2158, opened 2025-06-10
test: add unittest for DocumentationButton

sdobbert:
This adds a unittest for the component DocumentationButton.

### DIFF
--- a/src/components/DocumentationButton/index.spec.tsx
+++ b/src/components/DocumentationButton/index.spec.tsx
@@ -9,12 +9,6 @@ import {
 
 import DocumentationButton from '.';
 
-jest.mock('react-i18next', () => ({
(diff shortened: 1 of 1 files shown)


src/components/DocumentationButton/index.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+  cleanup
+} from '@testing-library/react';
+
+import DocumentationButton from '.';
+
+jest.mock('react-i18next', () => ({
+  useTranslation: jest.fn(() => ({
+    t: (key: string) => key
+  }))
+}));
```
dnlkoch, line 16: Is this mock needed at all? I would have assumed that it is handled by the [global mock](https://github.com/terrestris/shogun-gis-client/blob/main/jest.setup.js#L13-L22).

PR #2279, opened 2025-07-25
Allow uploading multiple file formats to the client

dnlkoch:
The current implementation of the `UploadDataModal` makes use of the GeoServer REST API to create a WMS based on selected input files. As this feature is mostly deactivated since we ususally don't want to create numerous services in it, this PR suggests to enable local client uploads instead.

Within this PR the range of supported formats is increased (basically all common formats are supported) and the user feedback is also improved by giving a short info about the detected data entries in the given dataset, e.g. the list of tables in a GeoPackage.

In addition it allows vector layers to be shown in the tree, enables some context menu entries (removal and zoom to extent) for them and makes them queryable with the info tool.

Peek 2025-07-25 16-31


dnlkoch:
Thanks for the review!

PR #2316, opened 2025-08-27
Adds logo link back to home

mholthausen:
This PR adds a logo link with a tooltip to return to the application's home page with a single click.


marcjansen:
Have we considered that a gis client might be running e.g. under a path/subdirectory in a URL; like e.g. `https://example.com/humpty/dumpty/gis-client/` ?

If understand the diff correctly the current implementation would always take a user to `https://example.com/`, but I can imagine situations in which `https://example.com/humpty/` or `https://example.com/humpty/dumpty/` or `https://example.com/humpty/dumpty/gis-client/` would be considered correct.

I think the link target should at least be configurable, even though often it might be correct with the current value.

dnlkoch:
Yes, that's a valid point. I think we would have three options here:

1. Reusing the existing global config `shogunBase` to link to the configured backend base url (that's usually pointing to the landing page).
2. Create a new global config. This would also allow to specify the url in client-only scenarios where `shogunBase` is set to `false`.
3. Create a new per application config to allow application specific urls.

I would favour option two, but please decide @mholthausen.

PR #2359, opened 2025-09-04
Move the current tool menu width to the store

dnlkoch:
This suggests to move the current tool menu width to the store. This is might become useful if other components or plugins need to change the width programmatically.


ahennr (approved):

dnlkoch:
Thanks for the review!

PR #2360, opened 2025-09-04
Pass plugin config to plugin component

dnlkoch:
This suggests to pass the plugin configuration (if any) to the plugin component. This enables the plugin to read the configuration without requesting it again.

The newly introduced type `PluginConfig` can be extended in plugins to match any given model, e.g.:

```tsx
import {
  ClientPluginComponentProps,
  PluginConfig
} from '@terrestris/shogun-gis-client/dist/plugin';

export type MyPluginConfig = PluginConfig & {
  myStringConfig?: string;
  myNumberConfig?: number;
};

export type MyPluginProps = ClientPluginComponentProps<MyPluginConfig>;

export const MyPlugin: React.FC<MyPluginProps> = ({
  config
}): JSX.Element => {
  return (<span>{config.myStringConfig}</span>);
};
```


ahennr (approved):
🔝 , very useful feature

dnlkoch:
Thanks for the review!

PR #2496, opened 2025-10-21
feat: refactor handling of shrinked map if drawer is opened

dnlkoch:
This suggests to update the handling of an opened `MapDrawer` and how this state is applied to the map and the map controls in size of width and position respectively:

- Add a new component `MapControl` to add react components as an OpenLayers control (and thus inside the `ol-overlaycontainer-stopevent` element). 
- Wrap the existing `MapToolbar` into the `MapControl`.
- Introduce a new prop `shrinkMapOnOpen` to the `MapDrawer` deciding if the map should be shrinked the remaining application size if the drawer is being opened
  - if `true` the map div will be shrinked and all controls inside the div will be moved accordingly
  - if `false` all `ol-controls` will be positioned accordingly

This makes the behaviour a bit more predictable and easier to override in plugins.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,7 +10,6 @@ import EditFeatureDrawer from './components/EditFeatureDrawer';
 import Footer from './components/Footer';
 import Header from './components/Header';
 import LayerDetailsModal from './components/LayerDetailsModal';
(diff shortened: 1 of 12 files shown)


dnlkoch:
Thanks for the review!

PR #2604, opened 2025-11-24
fix: wms version upper case

TreffN:
This PR changes the writting of `Version` to `VERSION` in the entered url in the `AddLayerModal`, as it is required by the [CapabilitiesUtil](https://github.com/terrestris/ol-util/blob/main/src/CapabilitiesUtil/CapabilitiesUtil.ts#L34) later on.

Please review


TreffN:
Handeled in v4.0.1 [base-util](https://github.com/terrestris/base-util), so this PR will be closed.

PR #2677, opened 2026-01-06
Merge branch 'main' into `next`

ahennr:
see title.

Beside, the default charsert is set to `utf-8`.

Plz review @terrestris/devs


dnlkoch:
Just did that [here](e25e6dd2b892a88c5230cec3242e8f9aa0a1be20) and also applied the fix the slice.

PR #2880, opened 2026-05-26
Adds ZoomFullExtent button

mholthausen:
This PR 

- adds support for a configurable `ZoomToExtentButton` in the map_toolbar made via the Tool Configuration of the shogun admin:

  `showZoomFullExtent: true` -> button is visible
  `showZoomFullExtent: false` or omitted -> button is hidden (default)
- fixes rspack dev startup by making React Refresh plugin import compatible with current `@rspack/plugin-react-refresh` exports. This resolves "ReactRefreshPlugin is not a constructor".

### DIFF
--- a/src/components/MapToolbar/index.tsx
+++ b/src/components/MapToolbar/index.tsx
@@ -8,6 +8,7 @@ import {
   faPlus,
   faMinus,
   faLocation,
+  faGlobe,
   faLocationPin
 } from '@fortawesome/free-solid-svg-icons';
 
@@ -26,6 +27,7 @@ import {
   GeoLocationButton
 } from '@terrestris/react-geo/dist/Button/GeoLocationButton/GeoLocationButton';
 import ZoomButton from '@terrestris/react-geo/dist/Button/ZoomButton/ZoomButton';
+import ZoomToExtentButton from '@terrestris/react-geo/dist/Button/ZoomToExtentButton/ZoomToExtentButton';
 
 import {
   useMap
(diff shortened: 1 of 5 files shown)


src/store/mapToolbar/index.tsx
```diff
@@ -5,7 +5,8 @@ import {
 
 const initialState = {
   visible: false,
-  showGeolocation: true
+  showGeolocation: true,
+  showZoomFullExtent: true
```
dnlkoch: The description states the button isn't visible by default. So is this intended here?

src/components/MapToolbar/index.tsx
```diff
@@ -57,12 +60,28 @@ export const MapToolbar: FC<MapToolbarProps> = (): JSX.Element => {
     }
   };
 
+  const initialCenter = map?.getView().getCenter();
+  const initialZoom = map?.getView().getZoom();
```
dnlkoch: Should we wrap the initial values into a memo (or a similiar solution) to protect them from being changed once the component is rerendered and the map's view has changed in the meantime?

src/store/mapToolbar/index.tsx
mholthausen: Yes, `showZoomFullExtent` in this slice was set to `true` for consistency with `showGeolocation`.
The effective visibility is controlled in `bootstrap.tsx` from `map_toolbar` tool config (provided via admin config).
So if the config is missing or set to `false`, the button remains hidden.

src/components/MapToolbar/index.tsx
mholthausen: Good catch. I’ve updated the implementation to derive the zoom target from `application.clientConfig.mapView` instead of the current map view state on re-render.

dnlkoch: It just occurred to me, but what happens if the center/zoom is provided by a permalink?

mholthausen: Permalink parameters can override the current map view on load, but this button is currently based on `application.clientConfig.mapView`. So it always returns to the configured initial view, not to a view provided by a permalink.

Is it intended for the button to take users to the initially loaded view, including the permalink, instead? This would be a different behavior and needs to be adjusted accordingly.

mholthausen: Then I would leave it as it is for now, and only the configured view will be used as the start view for this button.

dnlkoch (approved):

PR #2927, opened 2026-06-16
Enable automerge for non-major updates

dnlkoch:
As described in [here](https://docs.renovatebot.com/presets-default).


ahennr (approved):

ahennr:
Maybe presets ` ":automergeTesters"`, `":automergeDigest"` and `":automergeTypes"` could be added as well

dnlkoch:
Changed the config to make use of presets only.

PR #3003, opened 2026-08-04
test: improve element selection in e2e-tests

sdobbert:
This pull request updates the e2e-tests to use stricter role queries for button selection by adding the `exact: true` option to all `getByRole('button', { name: ... })` calls. This change ensures that the tests interact with the intended buttons only, reducing the risk of false positives.


dnlkoch (approved):
There are some linter warnings, otherwise +1

sdobbert:
> There are some linter warnings, otherwise +1
Thank you, I fixed the linting errors.

